PR #51, opened 2024-01-16
Add support for CUDA.jl 5.*

frapac:
We have to bump AMDGPU's version accordingly, as they both depend on GPUCompilers.jl.


frapac:
Closing in favor of #52

PR #57, opened 2024-02-20
Full Optimizer approach for MOI interface

sshin23:
With this approach,
```julia
using PowerModels, NLPModelsIpopt, ExaModels

@time pm = PowerModels.instantiate_model(
    joinpath(ENV["PGLIB_DEPOT"], "pglib_opf_case1354_pegase.m"),
    ACPPowerModel,
    PowerModels.build_opf
)
@time optimize_model!(
    pm,
    optimizer= ()->ExaModels.Optimizer(
        ipopt;
        linear_solver="ma27"
    )
)
```
and
```julia
using PowerModels, ExaModels, MadNLP, MadNLPHSL, MadNLPGPU, CUDA

@time pm = PowerModels.instantiate_model(
    joinpath(ENV["PGLIB_DEPOT"], "pglib_opf_case1354_pegase.m"),
    ACPPowerModel,
    PowerModels.build_opf
)
@time optimize_model!(
    pm,
    optimizer= ()->ExaModels.Optimizer(
        madnlp, CUDABackend()
    )
)

```
should work, though CUDA will complain that you're performing scalar indexing for the CUDA array (happening in non performance-critical part), which can be fixed in the future.

I'm not sure yet if the full optimizer approach would be the right approach here. 

## Upsides
- If we create an `Optimizer`, we can directly handle the interface with the solvers, so it will make interfacing with GPU-accelerated solvers (MadNLP) easier. 

## Downsides
- For each solver, extension packages need to be implemented `ExaModelsIpopt`, `ExaModelsMadNLP`, etc., which will eventually make the maintenance more difficult.

Thoughts? @odow @frapac @ccoffrin


sshin23:
One caveat though, is that we don't have interface to `MOI.Nonlinear` yet. We only have an interface to `MOIU.CachingOptimizer`'s expression tree. Thus, to make this as a `MOI.AutomatidDifferentiationBackend`, there should be some interface from JuMP that allows you to directly use `MOIU.CachingOptimizer` instead of hadling them via `MOI.Nonlinear`

sshin23:
@odow This is the approach taken in ExaModels.jl
https://github.com/exanauts/ExaModels.jl/blob/main/ext/ExaModelsMOI.jl#L28-L53

We build this nested `Bin`, which stores expression tree in `head` and stores data entries in `data`.

PR #105, opened 2024-08-04
Explicitly setting nln_nnzj

sshin23:
Fixes the issue coming from the introduction of `nnzj == lin_nnzj + nln_nnzj` check [NLPModels.jl ](https://github.com/JuliaSmoothOptimizers/NLPModels.jl/pull/468)


sshin23:
Fixed with the NLPModels v0.21.3 release

PR #112, opened 2024-10-11
Fix interface with NLPModels

frapac:
- add explicit return for the callbacks to fix `hess_structure` and `jac_structure` 
- fix type ambiguity in KernelAbstractions wrapper

xref #111


ext/ExaModelsKernelAbstractions.jl
```diff
@@ -366,13 +370,14 @@ function ExaModels.jprod_nln!(
...
     synchronize(m.ext.backend)
+    return Jv
 end
 function ExaModels.jtprod_nln!(
     m::ExaModels.ExaModel{T,VT,E},
     x::AbstractVector,
     v::AbstractVector,
     Jtv::AbstractVector,
-) where {T,VT,E<:KAExtension}
+) where {T,VT,E<:KAExtension{T,VT,NamedTuple}}
```
sshin23, line 380: Why do we want to specify the type parameters here?

src/nlp.jl
```diff
@@ -68,7 +68,7 @@ Base.show(io::IO, v::ConstraintAug) = print(
     io,
     """
 
-Constrant Augmentation
+Constant Augmentation
```
sshin23, line 72: Constraint

PR #113, opened 2024-11-05
[buildkite] Add CI builds on GPUs

amontoison:
@sshin23


sshin23:
test-self-hosted can be commented out

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,82 @@
...
+      queue: "juliagpu"
+      cuda: "*"
+    env:
+      EXAMODELS_TEST_CUDA: "TRUE"
+    command: |
+      julia --color=yes -e '
+        using Pkg
+        Pkg.add("CUDA")
+        Pkg.add("KernelAbstractions")
+        Pkg.add("NLPModels")
```
sshin23, line 16: why was this necessary?

.buildkite/pipeline.yml
```diff
@@ -14,6 +14,13 @@ steps:
         Pkg.add("CUDA")
         Pkg.add("KernelAbstractions")
         Pkg.add("NLPModels")
+        Pkg.add("JuMP")
+        Pkg.add("ForwardDiff")
+        Pkg.add("NLPModelsIpopt")
+        Pkg.add("NLPModelsJuMP")
+        Pkg.add("Percival")
+        Pkg.add("PowerModels")
+        Pkg.add("MadNLP")
```
sshin23, line 23: not sure if this is ideal. Maybe we can create a separate Project.toml file for testing?

.buildkite/pipeline.yml
amontoison: It depends on how you want to run the tests.
If you just want to use `runtests.jl`, we can modify it.
But with the current `pipeline.yml`, we can also run specific and longer tests that target specific GPUs without environment variables.

I did something similar in `Krylov.jl`, using buildkite to run GPU tests separately from CPU tests, which is recommended to avoid occupying the GPUs for too long, allowing other users and repositories to access them.

.buildkite/pipeline.yml
amontoison: The best solution is to create one `Project.toml` for each dependency. We can create a CI build for each extension if we do that.

sshin23: Can't we do, say, `Project.toml` on `/test` and have `Pkg.add("CUDA.jl")` and `Pkg.add("KernelAbstractions.jl")` before project environment instantiation?

amontoison: We can do that, it's what I do for KrylovPreconditioners.jl:
https://github.com/JuliaSmoothOptimizers/KrylovPreconditioners.jl/blob/main/.buildkite/pipeline.yml

sshin23: I think this approach is good. Can we do it in this way?

sshin23:
Thanks @amontoison for setting this up!

PR #124, opened 2024-12-05
Use Type instead of DataType

blegat:
I remember a talk of Jeff when he said that `DataType` should not be used and `Type` should be used instead


sshin23:
Thanks @blegat!

PR #137, opened 2025-05-16
minor fix (?) in quadrotor example

jbcaillau:
hi @sshin23 just to fix what looks like a (very) minor issue as `dt` should `T/N` (or `1/N`, depending in any case on `N` as, otherwise, changing the grid size implies changing the time horizon)

BTW: the mayer term in the cost is a bit strange (just looks like the reweighted final grid point of the Lagrange quadrature, but for the control):

```julia
    objective(c, 0.5 * Qf * (x[N+1, j] - d)^2 for (j, Qf, d) in itr2)
```

do you have a detailed ref. towards the problem?

cc: @amontoison

### DIFF
--- a/docs/src/quad.jl
+++ b/docs/src/quad.jl
@@ -4,12 +4,11 @@ function quadrotor_model(N = 3; backend = nothing)
 
     n = 9
     p = 4
-    nd = 9
     d(i, j, N) =
         (j == 1 ? 1 * sin(2 * pi / N * i) : 0.0) +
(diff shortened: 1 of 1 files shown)


frapac (approved):
LGTM!

sshin23:
Thank you, @jbcaillau for reporting this

jbcaillau:
> LGTM!

Looks Good To Me? Let’s Get To Merge… 🙂?

Issue #100, opened 2024-07-21
Unsafe use of MOI


mlubin:
The code below is accessing fields in a way that assumes `jm_cache` is a `MOI.Utilities.UniversalFallback`. This doesn't match the type of the argument (`MOI.ModelLike`) and could break in multiple ways. You should use the MOI API instead to query properties like variable bounds and starts.

https://github.com/exanauts/ExaModels.jl/blob/0ac40c25602229ecf8c65f5cf1ae70a13f1348bd/ext/ExaModelsMOI.jl#L58C10-L78

CC @blegat

PR #148, opened 2025-06-25
Add parameters

### LINKED ISSUE #147: Add support for model parameters in ExaModels
I am investigating the solution of optimization problems with parameters in ExaModels. Currently, the following syntax is breaking:
```julia
N = 10
params = rand(N)

core = ExaModels.ExaCore(Float64)
x = ExaModels.variable(core, N)
expr = ExaModels.constraint(core, x[i] + params[i] for i in 1:N; lcon=0.0, ucon=Inf)

```

The current solution is to pass the parameters directly in the iterator, like
```julia
N = 10
params = rand(N)

core = ExaModels.ExaCore(Float64)
x = ExaModels.variable(core, N)
expr = ExaModels.constraint(core, x[i]+p for (i, p) in zip(1:N, params); lcon=0.0, ucon=Inf)

```

However, that means that if we change the parameters, they are not easy to update in the model, as that means modifying the iterator inplace. 

I am wondering if we should add a new object in the expression tree to make parameters first-class citizen in ExaModels. A potential syntax would be 
```julia
N = 10
params = rand(N)

core = ExaModels.ExaCore(Float64)
x = ExaModels.variable(core, N)
p = ExaModels.parameters(params)
expr = ExaModels.constraint(core, x[i] + p[i] for i in 1:N; lcon=0.0, ucon=Inf)

```

An idea would be to implement a new node as: 
```julia
struct Parameter{I} <: AbstractNode
     i::I
end

```
and implements the same functions as for `Var`. 

The values of the parameters would be stored in a vector inside `ExaCore`. If we do that, it would be very easy to update the parameters. 

cc @klamike

klamike:
This PR adds parameters to ExaModels, which act like fixed variables to the user but to the solver look like scalars. This allows users to modify parameter values between solves without having to rebuild the model and without adding any extra variables and constraints. It also makes it possible to implement kernels for batches of problems.

Main changes:
- Changed the `AbstractNode` calls from two-argument `(i, x)` to three-argument `(i, x, θ)`. In fact, in general, everywhere `x` is passed, `θ` is now passed right after it. That is why there are so many tiny changes :) Note that all user-facing functions where the `m::ExaModel` is passed e.g. `obj`, `cons`, etc. are unchanged; they just take `x` and always use the current value of `m.θ`.
- Added `parameter`, `ParameterNode` and `ParameterSource`, which act like `variable`, `Var` and `VarSource` resp, and `set_parameter!` for updating the values.
- Updated the `ExaCore` and `ExaModel` to have an additional field `θ`, storing the initial/current value of the parameters.

Future work:
- Support `MOI.Parameter` in the MOI extension. I am hesitant to touch it due to #100
- Diff wrt parameters
- Evaluate performance compared to non-parametric
- Rename `Par`, `ParSource`, etc then rename `θ`

Closes #147 

cc @frapac


klamike:
I believe the test failure is unrelated, there seems to be a segmentation fault after the tests pass with oneAPI. I've verified that tests pass (and exit normally) on CUDA and AMD machines but don't have a oneAPI machine to debug that.

src/graph.jl
```diff
@@ -109,7 +115,7 @@ end
 
 @inline Base.getindex(n::ParSource, i) = ParIndexed(n, i)
 @inline Base.getindex(n::VarSource, i) = Var(i)
-
+@inline Base.getindex(::ParameterSource, i) = ParameterNode(i)
```
frapac, line 118: There is a bit a clash of naming between `ParSource` and `ParameterSource`, but this is not caused by this PR. I think at some point we should rename the object `ParSource` (this can be done later)

frapac (commented):
Good job, what you have implemented correspond exactly to what I had in mind when opening #147. I have a first round of comments, but the PR overall looks good to me. I think we need to discuss what would be an appropriate naming for the parameter `θ` (I would prefer avoiding using special character in the source code), but I am open for discussion.

docs/src/parameters.jl
```diff
@@ -0,0 +1,44 @@
...
+# ### Creating Parametric Models
+# 
+# Let's modify the example in [Getting Started](@ref) to use parameters. Suppose we want to make the penalty coefficient in the objective function adjustable:
+
+# First, let's create a core:
+using ExaModels, NLPModelsIpopt
+c_param = ExaCore()
+
+# Adding parameters is very similar to adding variables -- just pass a vector of values denoting the initial values.
+θ = parameter(c_param, [100.0, 1.0])  # [penalty_coeff, offset]
```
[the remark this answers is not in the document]
klamike: I'm not sure about the matrix/tensor, but I used similar logic to `variable` so it should support similar usages. The ACOPF test can cover this by using e.g. a matrix of parameters `[pd qd]`.

One thing to note is that currently users are forced to give starting values, similar to MOI.Parameter. That makes sense to me since it makes it impossible to forget to set the values. Of course, one can manually pass zeros if the value is not known at build time. So in a sense, there is no concept of "empty" parameters right now.

src/graph.jl
klamike: Agreed the clash is not ideal. I am not attached to any of the namings, open to change.

src/nlp.jl
```diff
@@ -131,6 +146,7 @@ Base.@kwdef mutable struct ExaCore{T,VT<:AbstractVector{T},B}
     nnzj::Int = 0
     nnzh::Int = 0
     x0::VT = convert_array(zeros(0), backend)
+    θ::VT = similar(x0)
```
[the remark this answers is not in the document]
klamike: I'll change it to `similar(x0, 0)`

As for the name, maybe `par0` or `p0` can work here? I just don't want to confuse it with `Par`/`ParSource`/`p` etc.

src/nlp.jl
```diff
@@ -362,6 +391,71 @@ function variable(
...
+    c::ExaCore,
+    param::Parameter,
+    values::AbstractArray
+)
+    Base.size(values) != param.size && throw(DimensionMismatch("Parameter size mismatch: expected $(param.size), got $(Base.size(values))"))
+    start_idx = param.offset + 1
+    end_idx = param.offset + param.length
+
+    converted_values = convert_array(values, c.backend)
+    c.θ[start_idx:end_idx] .= converted_values
```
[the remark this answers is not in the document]
klamike: updated in https://github.com/exanauts/ExaModels.jl/pull/148/commits/e172364f907b6c8b7d071c2ee57021bdf65983cf

klamike:
Documentation failure seems unrelated, it wants CUDA for the doctest: https://github.com/exanauts/ExaModels.jl/actions/runs/15903391894/job/44851938536?pr=148#step:5:11

src/nlp.jl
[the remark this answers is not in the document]
sshin23: I think `copyto!(@view(c.θ[start_idx:end_idx]), values)` should work.

src/graph.jl
sshin23: Agreed with renaming of `ParSource` et. al.

src/nlp.jl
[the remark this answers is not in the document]
sshin23: Is `p` taken? If we rename `Par` and `ParSource`, I don't think there will be confusion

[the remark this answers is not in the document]
sshin23: (agreed that it is better to avoid unicode characters)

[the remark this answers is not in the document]
klamike: I don't think it's taken in the struct, my point is more that `p` is used in the `show` methods/docs to represent the SIMD axis rather than model parameters.

sshin23:
In the long term, I think we'd need `ParametericNLPModels.jl`. @amontoison is there a package with a similar scope?

docs/src/parameters.jl
```diff
@@ -0,0 +1,44 @@
+# # [Parameters](@id parameters)
+# Parameters act like fixed variables. Internally, ExaModels keeps track of where parameters appear in the model, making it possible to efficiently modify their value without rebuilding the entire model.
+
+# ### Creating Parametric Models
+# 
+# Let's modify the example in [Getting Started](@ref) to use parameters. Suppose we want to make the penalty coefficient in the objective function adjustable:
```
klamike, line 6: I was hoping this would just work but alas... https://github.com/exanauts/ExaModels.jl/actions/runs/15905405697/job/44858854552?pr=148#step:5:79

How do I put a link to Getting Started there?

src/nlp.jl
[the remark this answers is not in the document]
sshin23: Then I think we can keep it as `θ` for now and open a separate PR for renaming

src/graph.jl
```diff
@@ -125,47 +131,51 @@ Par(iter::Type{T}, idx...) where {T<:NamedTuple} = NamedTuple{T.parameters[1]}(
...
+@inline (v::ParIndexed{I,n})(i, x, θ) where {I,n} = @inbounds v.inner(i, x, θ)[n]
 
-(v::ParIndexed)(i::Identity, x) = NaN # despecialized
-(v::ParSource)(i::Identity, x) = NaN # despecialized
+(v::ParIndexed)(i::Identity, x, θ) = NaN # despecialized
+(v::ParSource)(i::Identity, x, θ) = NaN # despecialized
 
 """
-    AdjointNode1{F, T, I}
+    AdjointNode1{F, T, S, I}
```
sshin23, line 149: Why does the first-order sensitivity have a different type than the function value here?

sshin23:
Thanks @klamike for the great contribution. @frapac and I talked about this a few days ago, but I was quite surprised that someone could implement this feature. Thanks for the great contribution again. Once this is merged, maybe we should talk about how to implement $\nabla_{xp} L$ and $(\nabla_{xp} L) v$...

amontoison:
> In the long term, I think we'd need `ParametericNLPModels.jl`. @amontoison is there a package with a similar scope?

@sshin23 To the best of my knowledge, I don't know any package with a similar scope.

docs/src/parameters.jl
klamike: Nope :( https://github.com/exanauts/ExaModels.jl/actions/runs/15905793204/job/44861326110?pr=148#step:5:31

src/graph.jl
sshin23: This is supposed to be a numerical value, e.g., `Float32` or `Float64`; there shouldn't be a difference even if we have parameters

klamike: Hmm actually it's due to parameter nodes there, it's `Int64` popping up for some reason...

klamike: Ah, I think I got it -- it's because I added types to the `ExaModels.AdjointNode1` function, making multiple dispatch reject it. Let me clean up the changes a bit and test it out locally.

sshin23: Do you have a commit hash with a single `T` version?

src/register.jl
```diff
@@ -138,6 +138,22 @@ macro register_bivariate(f, df1, df2, ddf11, ddf12, ddf22)
                 ExaModels.AdjointNode1($f, $f(x1, x2), $df2(x1, x2), d2)
             end
 
+            @inline function $f(
+                d1::ExaModels.ParameterNode,
```
sshin23, line 142: We need to ensure that $f is compiled for concrete data types. Should be
`d1::D1, ... ) where D1 >: ExaModels.ParameterNode`

src/graph.jl
klamike: Still working on it, would appreciate your help: c39b4ab5c70d895d9cc1fb21ebe7df9161078d1e

Getting this now:

```julia
  MethodError: Cannot `convert` an object of type ExaModels.ParameterNode{Int64} to an object of type ExaModels.Node2{typeof(*), ExaModels.ParameterNode{Int64}, Float64}
```

src/register.jl
sshin23: 👍

sshin23: This not resolved yet

klamike: Oh, do I need to add an `AbstractParameterNode` to put there instead?

sshin23: I meant https://github.com/exanauts/ExaModels.jl/commit/ba14dac355a5fab814f75c614bc7ecdd9d2e4974. Looks good now 👍

sshin23:
Looks complete once AC OPF test is implemented. @klamike, could you include a few timing results comparing native iterator vs parameter?

e.g., `constraint(core, c * x[i]^2 for (c,i) in itr)` vs `c = parameter(core, randn(10)); constraint(core, c[i]^2 for i in 1:N)`?

klamike:
```julia

julia> @benchmark constraint(core, ci * x[i]^2 for (ci,i) in itr) setup=(core=ExaCore())
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  15.333 μs …   4.088 ms  ┊ GC (min … max):  0.00% … 97.60%
 Time  (median):     36.166 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   49.542 μs ± 119.164 μs  ┊ GC (mean ± σ):  28.80% ± 14.64%

  ▂ █▁                                                          
  █▄██▅▃▂▂▂▂▂▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▂
  15.3 μs         Histogram: frequency by time          477 μs <

 Memory estimate: 450.03 KiB, allocs estimate: 51.

julia> @benchmark constraint(core, p[i]*x[i] for i in r) setup=(core=ExaCore();p=parameter(core,c))
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):   6.000 μs …   5.703 ms  ┊ GC (min … max):  0.00% … 98.87%
 Time  (median):     19.000 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   27.153 μs ± 100.625 μs  ┊ GC (mean ± σ):  28.18% ± 10.24%

   █▇                                                           
  ███▇▃▂▂▂▂▂▁▁▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂▂▁▂▂▂▂▂▂ ▂
  6 μs            Histogram: frequency by time          401 μs <

 Memory estimate: 289.34 KiB, allocs estimate: 34.

julia> @benchmark (p=parameter(core,c); constraint(core, p[i]*x[i] for i in r)) setup=(core=ExaCore())
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):   7.958 μs …  2.572 ms  ┊ GC (min … max):  0.00% … 98.16%
 Time  (median):     27.230 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):   35.736 μs ± 88.974 μs  ┊ GC (mean ± σ):  27.07% ± 12.17%

    █                                                          
  ▅▃█▇▃▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▁▁▂▂▂▂▂▂▂▂▂▂▂▂ ▂
  7.96 μs         Histogram: frequency by time         434 μs <

 Memory estimate: 385.41 KiB, allocs estimate: 37.

julia> @benchmark constraint(core, ci * x[i]^2 for (ci,i) in itr) setup=(core=ExaCore())
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  15.333 μs …   5.628 ms  ┊ GC (min … max):  0.00% … 98.81%
 Time  (median):     36.834 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   50.404 μs ± 119.612 μs  ┊ GC (mean ± σ):  26.58% ± 14.41%

    █▁                                                          
  ▇▃██▄▃▂▂▂▂▂▂▁▂▁▁▁▁▂▁▁▁▁▁▁▁▁▁▁▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▂
  15.3 μs         Histogram: frequency by time          506 μs <

 Memory estimate: 450.03 KiB, allocs estimate: 51.
```

Is this what you meant @sshin23 ? Going to look at how long it takes to solve 10k ACOPFs next

test/NLPTest/parameter_test.jl
```diff
@@ -0,0 +1,267 @@
...
+end
+
+function exa_ac_power_model_parametric(backend, filename; use_parameters=true)
+
+    data = parse_ac_power_data(filename, backend)
+
+
+    w = ExaModels.ExaCore(backend = backend)
+
+    pd = parameter(w, [b.pd for b in data.bus])
```
sshin23, line 69: This will cause a scalar indexing error. How about this?

```julia
parameter(w, map(d->d.pd, data.bus))
```

PR #150, opened 2025-06-26
Self-hosted runners added

sshin23:


sshin23:
I'll keep buildkite for now, but we can ditch it if the self-hosted runners work reliably

PR #154, opened 2025-06-29
Bump ExaModels version to 0.9.0

sshin23:


amontoison:
@sshin23 This version was never officially released.

Issue #132, opened 2025-03-31
Optimize the conversion between JuMP and ExaModels


joaquimg:
Given the model:
```julia
using ExaModels, JuMP, CUDA

model = Model()

@variable(model, ...)

@constraint(model, c1[i = ...], expr1... == 0.0)

@constraint(model, c2[i = ...], expr1... == 0.0)

@objective(model, Min, expr3...)
```
that would be converted with:
```julia
em = ExaModel(model; backend = CUDABackend())
```
It would be possible to use JuMP attributes to communicate part of the structure:
```julia
JuMP.set_attribute.(c1, ExaModels.ConstraintTag(), "c1")
JuMP.set_attribute.(c2, ExaModels.ConstraintTag(), "c2")
```
This way all the constraints in the `c1` list would be tagged with `"c1"` (analogous for `c2`).
Hence, the step of grouping constraints could be done only within the tag groups, which would minimize the work.

sshin23:
Hmm, that's an interesting idea. 

We still don't have a guarantee that each JuMP constraint will lead to a single type of ExaModel constraint. E.g., 
```julia
m = JuMP.Model()
@variable(model, x[1:10])
@constraint(model, [i=1:10], mod(i,2) == 0 ? sin(x[i]) : cos(x[i]) == 0)
```
will lead to two types of ExaModels constraint. 

Not sure how much difference this will make. We first convert the JuMP expression into an ExaModels expression
https://github.com/exanauts/ExaModels.jl/blob/main/ext/ExaModelsMOI.jl#L253-L410
and then classify them. Once they are converted into an ExaModel expression, classification is pretty quick. This tagging wouldn't help make conversion faster

joaquimg:
I admit that I do not fully understand the JuMP->Exa converter. My comments come from watching some presentations on ExaModels. I might have gotten some bits wrong.

About your example, tagging would be a user responsibility. In your example, the user would have to:
```julia
@constraint(model, c[i=1:10], mod(i,2) == 0 ? sin(x[i]) : cos(x[i]) == 0)
JuMP.set_attribute.(c[collect(i for i in 1:10 if mod(i,2) == 0)], ExaModels.ConstraintTag(), "c1")
JuMP.set_attribute.(c[collect(i for i in 1:10 if mod(i,2) != 0)], ExaModels.ConstraintTag(), "c2")
```

About converting and classifying, if conversion is the bottleneck, this would not help much.

[...]

PR #155, opened 2025-06-29
Update MOI extension

### LINKED ISSUE #100: Unsafe use of MOI

### LINKED ISSUE #130: ExaModels optimizer API for JuMP does not solve the correct model
As title says ExaModels optimizer API `ExaModels.IpoptOptimizer()` for JuMP does not solve the correct model and consequently has convergence issue. I found this by comparing many different AD backends. Please see discussion at [Discourse](https://discourse.julialang.org/t/examodels-optimizer-api-for-jump-has-convergence-issue/126032). There is detailed report and mwe.

klamike:
This PR updates `ExaModelsMOI` to use MOI safely. Closes #100, closes #130

Future work:
- Optimize the speed of conversion (function barriers?)
- Consider allowing to convert float type
- Consider supporting manual binning [ref](https://github.com/exanauts/ExaModels.jl/issues/132#issuecomment-2773026512)
- Consider implementing `MOI.Nonlinear.Evaluator` [ref](https://github.com/exanauts/ExaModels.jl/issues/132#issuecomment-2773404302)
- Better SIMD detection for nonlinear expressions (even basic ones like $\\ \sin(\sum_{i} x_{i})$ currently get flattened to $\sin(x_1+x_2+\cdots+x_n)$
- Detect easy cases like $\sum_{i} 1.0 * x_i\enspace\rightarrow\enspace\sum_{i} x_i\enspace$ or $\enspace 0.0 + \sum_{i} x_i\enspace\rightarrow\enspace\sum_{i} x_i$

### DIFF
--- a/ext/ExaModelsJuMP.jl
+++ b/ext/ExaModelsJuMP.jl
@@ -4,7 +4,7 @@ import ExaModels
 import JuMP
 
 function ExaModels.ExaModel(jm::JuMP.GenericModel{T}; options...) where {T}
-    return ExaModels.ExaModel(jm.moi_backend.model_cache; options...)
+    return ExaModels.ExaModel(JuMP.backend(jm); options...)
 end
 
 end # module ExaModelsJuMP
--- a/ext/ExaModelsJuMP.jl
+++ b/ext/ExaModelsJuMP.jl
@@ -4,7 +4,6 @@ import ExaModels
 import JuMP
 
 function ExaModels.ExaModel(jm::JuMP.GenericModel{T}; options...) where {T}
-    # FIXME: what is user passes `T` under `options`?
     return ExaModels.ExaModel(JuMP.backend(jm); T=T, options...)
 end
 
--- a/ext/ExaModelsJuMP.jl
+++ b/ext/ExaModelsJuMP.jl
@@ -4,7 +4,8 @@ import ExaModels
 import JuMP
 
 function ExaModels.ExaModel(jm::JuMP.GenericModel{T}; options...) where {T}
-    return ExaModels.ExaModel(JuMP.backend(jm); options...)
+    # FIXME: what is user passes `T` under `options`?
+    return ExaModels.ExaModel(JuMP.backend(jm); T=T, options...)
 end
 
 end # module ExaModelsJuMP
--- a/ext/ExaModelsMOI.jl
+++ b/ext/ExaModelsMOI.jl
@@ -7,18 +7,8 @@ const MOI = MathOptInterface
 const MOIU = MathOptInterface.Utilities
 const MOIB = MathOptInterface.Bridges
 
-const SUPPORTED_FUNC_TYPE = [
-    MOI.ScalarAffineFunction,
-    MOI.ScalarQuadraticFunction,
-    MOI.ScalarNonlinearFunction,
-    MOI.VariableIndex,
-]
-const SUPPORTED_SET_TYPE = [
-    MOI.GreaterThan,
-    MOI.LessThan,
-    MOI.EqualTo,
-    MOI.Interval,
-]
+const SUPPORTED_FUNC_TYPE = [MOI.ScalarAffineFunction, MOI.ScalarQuadraticFunction, MOI.ScalarNonlinearFunction, MOI.VariableIndex]
+const SUPPORTED_SET_TYPE = [MOI.GreaterThan, MOI.LessThan, MOI.EqualTo, MOI.Interval]
 const SUPPORTED_FUNC_TYPE_UNION = Union{SUPPORTED_FUNC_TYPE...}
 const SUPPORTED_SET_TYPE_UNION = Union{SUPPORTED_SET_TYPE...}
 
@@ -59,12 +49,7 @@ float_type(::Type{MOI.LessThan{T}}) where {T} = T
 float_type(::Type{MOI.EqualTo{T}}) where {T} = T
 float_type(::Type{MOI.Interval{T}}) where {T} = T
 
-function ExaModels.ExaModel(moim::MOI.ModelLike; backend = nothing, prod = false)
-    minimize = MOI.get(moim, MOI.ObjectiveSense()) === MOI.MIN_SENSE
-
-    variables = MOI.get(moim, MOI.ListOfVariableIndices())
-    con_types = MOI.get(moim, MOI.ListOfConstraintTypesPresent())
-
+function check_supported_and_deduce_T(con_types)
     set_float_types = Set()
     for (F,S) in con_types
         if !(F <: SUPPORTED_FUNC_TYPE_UNION)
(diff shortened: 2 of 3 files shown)


ext/ExaModelsMOI.jl
```diff
@@ -53,186 +44,248 @@ function _update_bin!(::BinNull, e, p)
...
+        !isnothing(ST) && push!(float_types, ST)
+    end
 
-function ExaModels.ExaModel(jm_cache::MOI.ModelLike; backend = nothing, prod = false)
+    obj_type = float_type(MOI.get(moim, MOI.ObjectiveFunctionType()))
+    !isnothing(obj_type) && push!(float_types, obj_type)
 
-    T = float_type(jm_cache.model)
-    minimize = jm_cache.model.objective.sense == MOI.MIN_SENSE
+    isempty(float_types) && error("Cannot determine float type") # FIXME: what if just nonlinear obj?
```
klamike, line 71: ref https://github.com/jump-dev/MathOptInterface.jl/issues/2776

ext/ExaModelsMOI.jl
```diff
@@ -53,186 +44,248 @@ function _update_bin!(::BinNull, e, p)
...
 end
-function _exafy_con(i, c::C, bin, offset; pos = true) where {C<:Real}
+function _exafy_con(i, c::C, bin, var_to_idx, con_to_idx; pos = true) where {C<:Real}
     e =
         pos ? ExaModels.ParIndexed(ExaModels.ParSource(), 1) :
         -ExaModels.ParIndexed(ExaModels.ParSource(), 1)
     bin = update_bin!(
         bin,
+        # FIXME: why do we need 0x1 here?
         ExaModels.ParIndexed(ExaModels.ParSource(), 2) => 0 * ExaModels.Var(1) + e,
```
klamike, line 238: Why do we need `0*x[1]` here?

ext/ExaModelsMOI.jl
```diff
@@ -53,186 +44,248 @@ function _update_bin!(::BinNull, e, p)
...
+    ucon = zeros(T, 0)
+    y0 = zeros(T, 0)
+    con_to_idx = Dict{MOI.ConstraintIndex, Int}()
+
+    for F in SUPPORTED_FUNC_TYPE
+        F <: MOI.VariableIndex && continue
+        FT = F <: MOI.ScalarNonlinearFunction ? F : F{T}
+        for S in SUPPORTED_SET_TYPE
+            ST = S{T}
+            cis = MOI.get(moim, MOI.ListOfConstraintIndices{FT,ST}())
```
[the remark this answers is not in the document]
klamike: It's an empty list so `exafy_con` ends up being no-op. Probably should add more explicit handling though

ext/ExaModelsMOI.jl
```diff
@@ -53,186 +44,248 @@ function _update_bin!(::BinNull, e, p)
...
 
-    for field in SUPPORTED_OBJ_TYPE
-        bin = exafy_obj(getfield(jobjs, field), bin)
+    c = ExaModels.ExaCore(T; backend = backend, minimize = minimize)
+
+    var_to_idx = copy_variables!(c, moim, T)
+    con_to_idx = copy_constraints!(c, moim, var_to_idx, T)
+    copy_objective!(c, moim, var_to_idx)
+
+    if return_maps
```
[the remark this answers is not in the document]
klamike: I think we need to return it to MOI in order for things like deleting variables/constraints to work. I'm pretty sure that would break with the current version. I plan to add a test for that

ext/ExaModelsMOI.jl
```diff
@@ -7,19 +7,10 @@ const MOI = MathOptInterface
...
-
-const SUPPORTED_CONS_TYPE =
-    [:moi_scalarnonlinearfunction, :moi_scalaraffinefunction, :moi_scalarquadraticfunction]
-const UNSUPPORTED_CONS_TYPE = [
-    :moi_vectoraffinefunction,
-    :moi_vectornonlinearfunction,
-    :moi_vectorquadraticfunction,
-    :moi_vectorofvariables,
-]
+const SUPPORTED_FUNC_TYPE = [MOI.ScalarAffineFunction, MOI.ScalarQuadraticFunction, MOI.ScalarNonlinearFunction, MOI.VariableIndex]
```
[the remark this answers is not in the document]
klamike: This is mostly me anticipating support for non-Float64 JuMP models e.g. JuMP.GenericModel{Float32}. But agreed this approach is kind of ugly, I will try to clean it up.

ext/ExaModelsMOI.jl
klamike: Ideally this whole function can be replaced by something like `JuMP.unsafe_backend` to get to the inner-most model which I think always has a concrete float type. Certainly over-complicating this part.

ext/ExaModelsMOI.jl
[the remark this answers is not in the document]
klamike: Maybe we can have a separate API like `to_exa`, then `ExaModel(moim::MOI.ModelLike) = first(to_exa(moim))`? I'm thinking LearningToOptimize will probably want the maps but not the whole `Optimizer` interface.

ext/ExaModelsJuMP.jl
```diff
@@ -4,7 +4,8 @@ import ExaModels
 import JuMP
 
 function ExaModels.ExaModel(jm::JuMP.GenericModel{T}; options...) where {T}
-    return ExaModels.ExaModel(jm.moi_backend.model_cache; options...)
+    # FIXME: what is user passes `T` under `options`?
+    return ExaModels.ExaModel(JuMP.backend(jm); T=T, options...)
```
klamike, line 8: Need to check this

ext/ExaModelsMOI.jl
```diff
@@ -559,4 +695,42 @@ function MOI.set(model::Optimizer, p::MOI.RawOptimizerAttribute, value)
...
 end
 
+
+_make_index_map(model::MOI.ModelLike, maps) = _make_index_map(model, maps[1], maps[2])
+function _make_index_map(model::MOI.ModelLike, var_to_idx, con_to_idx)
+    variables = MOI.get(model, MOI.ListOfVariableIndices())
+    map = MOI.Utilities.IndexMap()
+    for x in variables
+        map[x] = typeof(x)(var_to_idx[x][2]) # FIXME: this is wrong...
+    end
```
klamike, line 705: And this. I think we need to use [an offset ala POI](https://github.com/jump-dev/ParametricOptInterface.jl/blob/455541d2ed9870423ccf35684c5cf1e21370670f/src/ParametricOptInterface.jl#L21-L41)

ext/ExaModelsMOI.jl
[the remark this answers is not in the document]
frapac: this is a possible solution, but I am somewhat not entirely satisfied by it. If you agree, let's keep it that way for now, that's something we can fix in a future PR.

klamike:
The main extension should be pretty much final now. I do still plan to add the tests over the next few days

ext/ExaModelsMOI.jl
sshin23: I'm trying to remember why I needed this. Perhaps this was necessary to prevent errors for an empty objective

frapac (approved):

klamike:
Regarding #130, prior to this PR, the number of iterations was different between options (1) and (5) in the post (1 and 3 in the code). Now, the number of iterations and the iterates themselves match. However, not all the lines of the log match, namely the number of nonzeros in the Jacobian and Hessian are bigger when using ExaModels (4622 -> 6836 and 8198 -> 17108).

Since the iterates now match exactly, I think it's safe to say this PR closes #130.

test/JuMPTest/JuMPTest.jl
```diff
@@ -23,6 +23,106 @@ function jump_luksan_vlcek_model(N)
...
+    # cons = wcons.inner
+    # @test typeof(cons) <: ExaModels.ConstraintAug
+    # @test typeof(cons.f.f) <: Pair
+    # @test typeof(cons.f.f.second) <: ExaModels.Node2{typeof(*), ExaModels.ParameterNode{T1}, T2} where {T1<:ExaModels.ParIndexed,T2<:ExaModels.ParIndexed}
+    # @test typeof(cons.inner) <: ExaModels.ConstraintAug
+    # @test typeof(cons.inner.f.f) <: Pair
+    # @test typeof(cons.inner.f.f.second) <: ExaModels.Node2{typeof(*), ExaModels.Var{T1}, T2} where {T1<:ExaModels.ParIndexed,T2<:ExaModels.ParIndexed}
+    # @test typeof(cons.inner.inner) <: ExaModels.Constraint
+    # @test typeof(cons.inner.inner.f.f) <: ExaModels.Null{Nothing}
+    # @test typeof(cons.inner.inner.inner) <: ExaModels.ConstraintNull
```
klamike, line 78:
```suggestion
    jm = JuMP.Model()

    JuMP.@variable(jm, x[1:N])
    JuMP.@variable(jm, p in JuMP.Parameter(1.0))
    JuMP.@constraint(jm, sum(x) == p)
    JuMP.@objective(jm, Min, sum(x))

    em = ExaModel(jm)
    @test only(em.meta.lcon) == only(em.meta.ucon) == 0.0
    @test only(em.θ) == 1.0
    wcons = em.cons
    @test typeof(wcons) <: ExaModels.ConstraintAug
    @test typeof(wcons.f.f) <: ExaModels.Null
    cons = wcons.inner
    @test typeof(cons) <: ExaModels.ConstraintAug
    @test typeof(cons.f.f) <: Pair
    @test typeof(cons.f.f.second) <: ExaModels.Node2{typeof(*), ExaModels.ParameterNode{T1}, T2} where {T1<:ExaModels.ParIndexed,T2<:ExaModels.ParIndexed}
    @test typeof(cons.inner) <: ExaModels.ConstraintAug
    @test typeof(cons.inner.f.f) <: Pair
    @test typeof(cons.inner.f.f.second) <: ExaModels.Node2{typeof(*), ExaModels.Var{T1}, T2} where {T1<:ExaModels.ParIndexed,T2<:ExaModels.ParIndexed}
    @test typeof(cons.inner.inner) <: ExaModels.Constraint
    @test typeof(cons.inner.inner.f.f) <: ExaModels.Null{Nothing}
    @test typeof(cons.inner.inner.inner) <: ExaModels.ConstraintNull
```
when #158 is merged this should pass

sshin23 (approved):

PR #176, opened 2025-08-06
Fix nesting structs in generators.

hfytr:
This pr allows nesting of structures in generators passed to constraint, objective, etc. the following sample is fixed by the pr:
```julia
struct A
    foo::Int
end
core = ExaCore(Float64; backend = nothing)
v = variable(core, 1)
a = [A(1)]
c = constraint(core, v[i] < i * a.foo for (i, a) in enumerate(a)) # previously would error here, with no method indexed_iterate for ParIndexed
```

### DIFF
--- a/src/graph.jl
+++ b/src/graph.jl
@@ -117,6 +117,10 @@ end
 @inline Base.getindex(n::ParSource, i) = ParIndexed(n, i)
 @inline Base.indexed_iterate(n::ParSource, idx, start = 1) = (ParIndexed(n, idx), idx + 1)
 
+@inline Base.getproperty(v::ParIndexed{I, n}, s::Symbol) where {I, n} = ParIndexed(v, s)
+@inline Base.getindex(v::ParIndexed{I, n}, i) where {I, n} = ParIndexed(v, i)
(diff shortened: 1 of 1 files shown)


sshin23:
@hfytr could you add a test for this?

hfytr:
Is it ok if i just add the static case opf from ExaModelsPower?

hfytr:
@sshin23 I also implemented show and string for Node types. let me know if you want this removed.

src/graph.jl
```diff
@@ -92,6 +92,16 @@ struct Node1{F,I} <: AbstractNode
     inner::I
 end
 
+import Base.string, Base.show
+
+Base.string(n::Node1{F,I}) where {F,I} = "($(string_f(F)) $(string(n.inner)))"
```
sshin23, line 97: why do we do `(sin x)` instead of `sin(x)`?

sshin23: We should prettyprint `Var` and `Par` as well

sshin23:
Intel error is unrelated

sshin23:
Thanks, @hfytr. Great work!

PR #191, opened 2025-11-18
Support for ObjectiveSense

blegat:
Otherwise I get an error when changing the objective function as it think the issue is that `ObjectiveSense` is not supported, not that the model should be wiped out and `copy_to` should be called again

### DIFF
--- a/ext/ExaModelsMOI.jl
+++ b/ext/ExaModelsMOI.jl
@@ -700,6 +700,9 @@ function MOI.supports_constraint(
 )
     return true
 end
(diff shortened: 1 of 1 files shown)


klamike:
Isn't a corresponding setter required here for MOI to actually change the sense? Something like

```julia
function MOI.set(
    model::Optimizer,
    ::MOI.ObjectiveSense,
    sense::MOI.OptimizationSense,
)
    model.model.meta.minimize = isequal(sense, MOI.MIN_SENSE)
    return
end
```

michel2323:
@blegat Can you rebase to rerun the tests?

PR #200, opened 2025-11-25
Track variable bounds for ExaModels.Optimizer

### LINKED ISSUE #199: ExaModels.Optimizer fails when variables have bounds
```julia
julia> m = Model(() -> ExaModels.Optimizer(madnlp, CUDABackend()))
A JuMP Model
├ solver: madnlp running with ExaModels
├ objective_sense: FEASIBILITY_SENSE
├ num_variables: 0
├ num_constraints: 0
└ Names registered in the model: none

julia> @variable m x ≥ 1
x

julia> @objective m Min x^2 + x + 1
x² + x + 1

julia> optimize!(m)
ERROR: KeyError: key MathOptInterface.ConstraintIndex{MathOptInterface.VariableIndex, MathOptInterface.GreaterThan{Float64}}(1) not found
Stacktrace:
```

The reason is we skip adding variable bounds to `con_to_idx`. Note this is not an issue with `ExaModel(m)` since that path doesn't create the `IndexMap`

klamike:
Fixes #199. Turns out the `Model(() -> ExaModels.Optimizer)` was not being tested, only `ExaModel(model)`. The added test fails before this change.

### DIFF
--- a/test/JuMPTest/JuMPTest.jl
+++ b/test/JuMPTest/JuMPTest.jl
@@ -307,7 +307,6 @@ function runtests()
                             m = WrapperNLPModel(ExaModel(jm; backend = backend))
                             result = ipopt(m; print_level = 0)
 
-                            # NOTE: assumes IndexMap is identity!
                             @test sol ≈ result.solution atol = 1e-6
                         end
                     end


test/JuMPTest/JuMPTest.jl
```diff
@@ -292,12 +292,21 @@ function runtests()
...
+                    @test sol ≈ sol2 atol = 1.0e-6
+                    @test dsol ≈ dsol2 atol = 1.0e-6
 
                     for backend in BACKENDS
                         @testset "$backend" begin
                             m = WrapperNLPModel(ExaModel(jm; backend = backend))
                             result = ipopt(m; print_level = 0)
 
+                            # NOTE: assumes IndexMap is identity!
                             @test sol ≈ result.solution atol = 1e-6
```
klamike: Maybe we should just check the objective value here?

The comment is actually not quite true. It assumes the keys are in the same order, not necessarily identity..

klamike:
Github's macOS runners + LTS + OpenCL seems to be a bad combo https://github.com/JuliaGPU/OpenCL.jl/pull/406. We should probably reduce the max CI time... and I should push less often :wink:

klamike:
Failure seems unrelated, probably ExaModelsOneAPI needs some more type piracy

amontoison (approved):

PR #210, opened 2026-01-23
Add TwoStageExaModel for two-stage optimization

michel2323:
Adds `TwoStageExaModel` for two-stage stochastic optimization with block-arrowhead structure.

**Two modes:**
- Default: per-scenario builder `(c, d, v, θ, i)` - simple API
- SIMD (`simd=true`): single compiled pattern `(c, d, v, θ, ns, nv, nθ)` - max GPU efficiency

**Structure:** Variables `[v₁; v₂; ...; vₛ; d]`, Constraints `[c₁; c₂; ...; cₛ]`

**API:** `TwoStageExaModel(build, nd, nv, ns, θ_sets; simd=false, backend=nothing)`

--> [preview](https://exanauts.github.io/ExaModels.jl/previews/PR210/two_stage) <--

### DIFF
--- a/docs/src/two_stage.jl
+++ b/docs/src/two_stage.jl
@@ -164,7 +164,7 @@ println("Optimal objective: ", round(result3.objective, digits = 4))
 
 # ## Accessing the Underlying Model
 # For advanced use cases, you can access the underlying `ExaModel`:
-inner_model = ExaModel.get_model(model)
+inner_model = ExaModels.get_model(model)
 println("\nUnderlying model type: ", typeof(inner_model))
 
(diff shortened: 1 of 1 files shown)


src/two-stage.jl
```diff
@@ -0,0 +1,552 @@
...
+
+Evaluate all constraints.
+Output: c_global ∈ ℝ^{ns*nc}
+"""
+function cons!(
+        model::TwoStageExaModel,
+        x_global::AbstractVector,
+        c_global::AbstractVector
+    )
+    cons_nln!(model.model, x_global, c_global)
```
amontoison:
```suggestion
    cons!(model.model, x_global, c_global)
```

PR #223, opened 2026-02-20
Check dimension and throw error

sshin23:


sshin23:
Will revisit this later

PR #226, opened 2026-02-20
New two-stage API

### LINKED ISSUE #221: New two-stage API
```julia
using ExaModels

ns = 10

c = ExaCore(scenario = true)
x = variable(c, 1:5; scenario = [1, 3, 2,4 5])
c1 = constraint(c, x[i] == 1 for i in 1:5; scenario = [1, 3, 2,4 5])
```

sshin23:
close #221

### DIFF
--- a/src/tags.jl
+++ b/src/tags.jl
@@ -22,3 +22,4 @@ function TwoStageExaCore(args...; backend = nothing, kwargs...)
     )
 end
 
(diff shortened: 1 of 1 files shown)


amontoison:
@sshin23 We didn't updated all the tests and we currently don't run any of them.

PR #230, opened 2026-02-25
LinearAlgebra extension for ExaModels nodes

michel2323:
- Adds a new package extension ExaModelsLinearAlgebra that enables LinearAlgebra operations (dot, norm, matrix multiply, LU solve) on  ExaModels Null and AbstractNode types                                                                                                       
- Implements null node pass-throughs, scalar arithmetic with zero/one elimination, and matrix/vector operations


jbcaillau:
> Jean-Baptiste will check that the current version of the branch is solving all his issues. I am waiting his green light 🚥

@amontoison green light

amontoison (approved):

sshin23:
I'm currently implementing AOT compilation, and we need to be clever to allow AOT compilation for this. I'm hoping that using StaticArrays can help. Could you point me to the use case of this? @michel2323 @jbcaillau @amontoison?

sshin23:
Right now, we don't have test for `constraint(core, A * x )` sort of use case

sshin23:
Would this be the kind of use case?

```julia
x = variable(core, 2, 10)
u = variable(core, 2, 10)
f(x, u) = [x[1]^2 + u[1], sin(x[2]) * u[2]]
contraint(core, x[:,i+1] - f(x[:,i], u[:,i]) for i = 1:9)
```

pulsipher:
Would one possible use case be embedding dense NN layers as algebraic constraints via something like `A * x` instead of using `constraint!` to handle the sums? For instance, the use of `constraint!` in MathOptAI (https://github.com/lanl-ansi/MathOptAI.jl/blob/main/ext/ExaModels/Affine.jl) could be replaced with a linear algebra operation which I believe would be a significant simplification.

sshin23:
The compilation time for this part of the code
https://github.com/lanl-ansi/MathOptAI.jl/blob/main/ext/ExaModels/Affine.jl#L25-L29
will scale very poorly as n increases. To avoid the outer loop, it will probably be necessary to use a double iteration over i and j within the generator for `constraint( )`. Supporting `A * x` as implemented in this extension would likely make the compilation blow-up even worse.

Implementing `A * x` itself may still be manageable, but what happens if the user wants something like `tanh.(A * tanh.(B * tanh.(C * x)))`? At that point, we would essentially be re-implementing Torch. I am therefore inclined not to support generic vector syntax.

ext/ExaModelsLinearAlgebra.jl
```diff
@@ -0,0 +1,517 @@
...
+function ExaModels.constraint(
+        c::ExaModels.ExaCore{T}, v::VecExaNode;
+        start = zero(T), lcon = zero(T), ucon = zero(T), kwargs...,
+    ) where {T}
+    n = length(v)
+    _get(x::AbstractVector, i) = x[i]
+    _get(x, _) = x
+    c1 = nothing
+    for i in 1:n
+        c1 = ExaModels.constraint(
```
sshin23, line 509: @michel2323 potentially type instability

ext/ExaModelsLinearAlgebra.jl
```diff
@@ -0,0 +1,517 @@
...
+# ============================================================================
+# Section C: Type aliases, promotion, and adjoint for nodes
+# ============================================================================
+
+const ExaNode = Union{
+    ExaModels.AbstractNode,
+    ExaModels.AbstractAdjointNode,
+    ExaModels.AbstractSecondAdjointNode,
+}
+const VecExaNode = AbstractVector{<:ExaNode}
```
sshin23, line 106: should be careful with non-uniform length

sshin23 (commented):
Sorry for the late comment. Just took a look at the PR. @michel2323 @amontoison @jbcaillau Could you point me to the place where these are used?

PR #233, opened 2026-02-27
ExaVector and ExaMatrix

amontoison:
Modifications of @michel2323


sshin23:
As discussed today, I want to be careful with supporting vectors/matrices of ExaModel expressions. Closing this

PR #236, opened 2026-03-02
Use Adapt.jl for backend array conversion

michel2323:
@sshin23 , This is equivalent to https://github.com/exanauts/ExaModels.jl/pull/235, but uses `Adapt`.


klamike:
I gave metal support a shot a while ago. I don't think it's complete but maybe it can be useful for you: https://github.com/exanauts/ExaModels.jl/compare/main...klamike:ExaModels.jl:mk/metal

sshin23:
We have some `NaN <: Float64` in our code that will cause some issues
https://github.com/search?q=repo%3Aexanauts%2FExaModels.jl%20NaN&type=code

sshin23 (approved):

michel2323:
Maybe waiting for this https://github.com/JuliaGPU/OpenCL.jl/pull/423 so we can remove OpenCL too.

michel2323:
The OpenCL PR takes too long. Merging.

PR #237, opened 2026-03-02
Add extension to GenOpt

blegat:
To add GenOpt support to ExaModels, the current approach in https://github.com/blegat/GenOpt.jl/blob/main/src/examodels.jl was to copy-paste the MOI wrapper of ExaModels and then hack it.
In this PR, I'd like to see if it's possible to extend the MOI wrapper here because since it's heavily dependent on the internals of the MOI wrappers, it will be easier to maintain it in the same repo as the MOI wrapper


Project.toml
```diff
@@ -36,6 +38,7 @@ ExaModelsSpecialFunctions = "SpecialFunctions"
 
 [compat]
 Adapt = "4"
+GenOpt = "0.1"
```
blegat:
```suggestion
GenOpt = "0.2"
```

src/nlp.jl
```diff
@@ -49,7 +49,7 @@ end
 Base.show(io::IO, s::Expression) = _show_expression(io, s)
 function _show_expression(io::IO, s::Expression)
     expr = try
-        _expr_string(s.f(ParSource()))
+        _expr_string(s.f(DataSource()))
```
blegat, line 56: This was a leftover from https://github.com/exanauts/ExaModels.jl/pull/234

sshin23:
Sry-the Metal issue probably was me closing the ci laptop lid

src/wrapper.jl
```diff
@@ -0,0 +1,166 @@
...
+function exafy_extension_obj_arg end
+exafy_extension_obj_arg(m, var_to_idx) = nothing
+
+"""
+    op(s::Symbol)
+
+Map a Symbol to the corresponding Julia function. Used by both ExaModelsMOI
+and ExaModelsGenOpt for expression tree conversion.
+"""
+function op(s::Symbol)
```
sshin23: Why was this moved from ExaModelsMOI.jl?

sshin23 (commented):
I’m fine with whatever changes to ExaModelsMOI.jl are needed to make GenOpt work, though I’m not sure whether ExaModelsGenOpt.jl should be stored in the ExaModels.jl repository. Wouldn’t it make more sense to keep it in GenOpt.jl instead?

sshin23:
Looks like there are no type piracy concerns

blegat:
> I’m fine with whatever changes to ExaModelsMOI.jl are needed to make GenOpt work, though I’m not sure whether ExaModelsGenOpt.jl should be stored in the ExaModels.jl repository. Wouldn’t it make more sense to keep it in GenOpt.jl instead?

Good question. My feeling is that it might be easier to maintain here because ExaModelsMOI.jl and ExaModelsGenOpt.jl are intertwined in a non-trivial way. If we added an ExaModels extension to GenOpt instead, it would mean that what we put in `src/wrapper.jl` should be a well-defined public API and we should come up with precise behavior that ExaModelsMOI.jl should keep in order for the ExaModels extension to GenOpt to keep working, and even add tests here to make sure ExaModelsMOI.jl don't break that. If it does and it is needed, then it will need to trigger a breaking release for ExaModels. On the other hand, with the current design having ExaModelsMOI.jl and ExaModelsGenOpt.jl both in ExaModels, any modifications of ExaModelsMOI.jl that don't break the tests in `test/GenOptTest` is fine, we don't need to be extra careful when modifying ExaModelsMOI.jl. Of course, then we would need to modify ExaModelsGenOpt.jl when GenOpt makes a breaking release but as ExaModelsGenOpt.jl only uses what's already public API from GenOpt, we won't need to make more breaking releases because the extension is in ExaModels. So it seems overall easier to maintain here. Of course, that puts the maintainance on your side but I can take care of updating it whenever GenOpt makes a breaking release :)

blegat:
The runic failure is just `Refusing to check out fork pull request code from a 'pull_request_target' workflow`

PR #242, opened 2026-03-03
Add Metal support

sshin23:


sshin23:
This was much more work than I originally thought, but I think it was worth it, not because Metal is useful, but because we can ensure that the Float32 calculation is done purely with Float32. 

Metal doesn't run on GitHub CI. I locally tested, and the test passes. I'll configure the runner on my old laptop later...

michel2323:
Nice. It should run on the CI. I'll try to fix that.

Edit: You were right. VMs on GH don't allow access to the GPU. How sad.

michel2323:
I think the MacOS runner for Metal should be labeled as `aarch64` instead of `X64`. I switched Julia to use the native aarch64 binaries. I doubt Metal works with x64 emulation. We should also drop x64 for the GitHub runner on macOS. That might be why it sometimes hangs.

sshin23:
@michel2323 fyi, I set up my old laptop as a ci

michel2323:
> @michel2323 fyi, I set up my old laptop as a ci

Can you install an arm runner with `aarch64` architecture? Or is this an Intel Mac? Metal is not well supported on Intel x64.

sshin23:
This is an m1 macbook air, so it is aarch64

michel2323:
The Github Actions self-hosted runner is still x86. So it crashes somehow since it runs in Rosetta I suppose.

Issue #239, opened 2026-03-03
Add @register_multivariate for n-argument custom operators


andrewrosemberg:
**Current behavior**

ExaModels provides `@register_univariate` and `@register_bivariate` for registering custom scalar operators with user-supplied derivatives. There is no equivalent for functions of 3 or more scalar arguments.

**New feature**

A `@register_multivariate(f, n, grad!, hess!)` macro (or equivalent) that registers an `n`-argument scalar function with explicitly provided gradient and Hessian:

```julia
# f(x₁, ..., xₙ) -> scalar
# grad!(g, x₁, ..., xₙ) fills the n-vector of partial derivatives
# hess!(H, x₁, ..., xₙ) fills the n×n lower-triangular Hessian
```

**Personal Motivation**

My primary use case is embedding external physics "black-box" simulators (e.g., MuJoCo MJX, Dojo) into ExaModels problems: `x_{t+1} - f(x_t, u_t) = 0` with available Jacobians and Hessians.

klamike:
It may be better to support this kind of application via a MOI.VectorNonlinearOracle-type approach, rather than an operator registration approach. That would avoid having to support n-arity nodes in ExaModels since we can skip the expression building/AD entirely for these constraints. There can be an API like

```julia
oracle_constraint(n_input, n_output, f, jac, jac_structure; hess, hess_structure, jprod, jtprod, hprod)
```

which ExaModels would store as pointers to the user functions, and maintain counters to figure out how to slice the combined buffers the users expect us to fill.

andrewrosemberg:
> It may be better to support this kind of application via a MOI.VectorNonlinearOracle-type approach, rather than an operator registration approach. That would avoid having to support n-arity nodes in ExaModels since we can skip the expression building/AD entirely for these constraints. There can be an API like

Started https://github.com/exanauts/ExaModels.jl/pull/243

sshin23:
I wonder if it would make sense to do this even at the `NLPModels.jl` level. What would be the advantage of doing this along with `ExaModels`?

klamike:
I think it is more of a modeler concern... NLPModels shouldn't care about the implementation of `cons`, if it comes from an oracle or not. Are you suggesting we have a `cons_oracle` similar to `cons_lin` and `cons_nln`?

sshin23:
[...]

PR #243, opened 2026-03-04
Add vector nonlinear oracle

andrewrosemberg:


sshin23:
Thanks for the contribution @andrewrosemberg is this ready for review?

andrewrosemberg:
> Thanks for the contribution @andrewrosemberg is this ready for review?

yes :)

michel2323:
This is also needed for MathOptAI.jl "gray box" support. I'll take a look and make sure it works with that too.

michel2323:
After digging into this, I came to the conclusion that we need `@register_multivariate` https://github.com/exanauts/ExaModels.jl/issues/239 . 

In case of `h(x) = g(f(x))`, where `g` is an ExaModel expression and `f` the oracle, the oracle `f` will always assemble its full Hessian/Jacobian. In the case where the oracle is a NN, that would be dense. However, the whole model `h` might still be sparse.

So instead of exposing `hess!` and `jac!`, we actually need to expose `hvp!` and `jvp!`.

In JuMP, they have wanted to add this for a while for the oracle. However, it does not seem so urgent, as the current solution works for small models. For ExaModels, I think `hess!` and `jac!` is not really a use case as it breaks the whole SIMD. What do you think?

andrewrosemberg:
> What do you think?

Happy with adding the optional hvp! and jvp!. Will try to add it later in the week and test it with my robotics example

andrewrosemberg:
@michel2323 , I am having difficulty testing jvp!/vjp!/hvp! with madnlpgpu. can't seem to figure out if they are being used or not. I tried lbfgs mode (since it explicitly uses jacobian vector product) but it is only compatible with CPU AFAIK. How can I take advantage of these vector-product callbacks in madnlp GPU?

michel2323:
I'm gonna take a look. I have to say, this is a bit challenging for me, too. I will take a look now. I've added you to the repository, and I'll move the branch over so we can push together.

michel2323:
This adds an oracle implemented as slack variables. I'm still not fully happy with it, but I think it's a first step. I think there is still an issue with not exploiting sparsity when embedding a dense oracle (e.g., an NN) into a sparse model. In that case, I think a reduced space approach might be useful.  However, that currently works through MathOptAI.jl. So, not sure if this is needed for general operators.

@sshin23, @amontoison, if you have time,  can you also take a look? The API is illustrated in an example `docs/src/oracle.jl`.

amontoison:
@michel2323 I can have a look this weekend. Ping me again if I forget.

sshin23:
Thanks for the contribution @andrewrosemberg. It took some time for us to make some major changes in ExaModels syntax. I'll rebase it to main first, given that lots of ExaModels internals have changed

michel2323:
Cleaned up a bit and rebased.

michel2323:
This PR is converging and was far more complicated than anticipated, but I think having a well designed oracle in ExaModels for coupling with other GPU codes is important.

Highlights:
* A nonlinear vector oracle for constraints
* A scalar nonlinear oracle for objective
* Both have GPU backend support and avoid any allocations during optimization loop
* Both support a CPU fallback in case only the ExaModels model is run on the GPU
* Full support for matrix-free jvp and hvp evaluation (!)

In summary, we have three oracle flavors :                                                                                                                         
                                                                                                                                                         
  1. `VectorNonlinearOracle`: A block of `ncon` nonlinear constraints. Adds new rows to the model. Registered via `constraint(core, oracle`). Callbacks: `f!(residuals)`, plus either explicit `jac!` / `hess!` or matrix-free `jvp!` / `vjp!` / `hvp!`. This is what `embed_oracle` is built on.                               
  2. `ScalarNonlinearOracle`: A single scalar term added to the objective. No new constraints, no Jacobian. Registered via `objective(core, oracle)`. Callbacks: `f(value)`, `grad!`, optional `hvp!`.                                                                                                            
  3. `OracleEvaluator`: Augments pre-existing constraint rows declared via `add_con`. No new rows; just fills values into rows that already exist. Registered via `add_eval(core, cons_tuple, vars_tuple, f!; …)` or the macro form `@add_con!(core, (c,…), (x,…), f!; …)`. Same callback shape as the vector oracle (`f!`, `jac!`/`hess!` or `jvp!`/`vjp!`/`hvp!)`, but operating on local concatenated index spaces.

@sshin23 @andrewrosemberg Feel free to take a look!

PR #244, opened 2026-03-05
Two-stage model

michel2323:
This is just a placeholder and WIP. I'm trying to get an overview what is needed to get a parallel two-stage solver added to MadNLP.jl using CUDSS.jl.


michel2323:
Ok. I think this light layer is enough. I got a roundtrip with this and batched CUDSS for the scenarios.

src/two_stage.jl
```diff
@@ -0,0 +1,84 @@
+struct TwoStageExaModel{T, M <: ExaModel{T}}
+    model::M
+    ns::Int
+    nv::Int
+    nd::Int
+    nθ::Int
+    nc_per_s::Int
+end
+
+function TwoStageExaModel(
```
sshin23: @michel2323 I don't see why this builder function can be useful for the users. It's likely that the user wouldn't like the "combined" scenario variable `v`, since they may prefer to work with different partitioning within `v`. Why can't we allow the user to directly build the two-stage model?

sshin23: I also wonder if `TwoStageExaModel` can be implemented by allowing some sort of extension of `ExaModel` as opposed to building on top of it, so that it can be just treated as an `AbstractNLPModel`

sshin23: I think the following syntax might be more useful for the user
```julia
x = variable(core, 1:10) # creates a design variable
y = variable(core, 1:10, EachScenario())) # creates 10-dimensional variable for each scenario
c = constraint(core, x[i] + y[i] for i=1:10, EachScenario()) # creates constraint for each scenario
```

michel2323: Sounds good to me. Let me get that done and work with MadNLP.

src/two_stage.jl
```diff
@@ -0,0 +1,70 @@
+# --- Index accessors (ExaModel with TwoStageTags) ---
+"""
+    scenario_var_indices(model::ExaModel, s::Int)
+
+Return indices of variables belonging to scenario `s` (from `TwoStageTags`).
+"""
+function scenario_var_indices(model::ExaModel, s::Int)
+    tags = model.tags::TwoStageTags
+    return findall(==(s), Array(tags.var_scenario))
```
sshin23: I think we should define `TwoStageExaModel` and only define these functions for  `TwoStageExaModel`

sshin23: Is `Array` necessary?

src/tags.jl
```diff
@@ -16,11 +35,51 @@ function TwoStageExaCore(args...; backend = nothing, concrete = Val(false), kwar
...
+# --- EachScenario variable method ---
+"""
+    variable(core::TwoStageExaCore, n, EachScenario(); kwargs...)
+
+Create `n` variables per scenario (total `ns * n` variables), automatically
+tagged with their scenario index. Design variables (shared) are created
+with the standard `variable(core, n)` call (tagged with scenario 0).
+"""
+function variable(
+    c::C, n::Integer, ::EachScenario;
```
sshin23: We allow `variable(c, ns...)`, where `ns...` can be either an integer or a unit range.  We should make this use new syntax `add_var`

src/tags.jl
```diff
@@ -16,11 +35,51 @@ function TwoStageExaCore(args...; backend = nothing, concrete = Val(false), kwar
...
+
+# --- EachScenario constraint methods ---
+"""
+    constraint(core::TwoStageExaCore, gen::Base.Generator, EachScenario(); kwargs...)
+
+Create constraints from a generator and automatically tag them per scenario.
+The total number of constraints must be divisible by `ns`. Constraints are
+assigned to scenarios in order: first `nc_per_s` go to scenario 1, etc.
+"""
+function constraint(
```
sshin23: Here, we should support all syntax of `add_con`

src/tags.jl
```diff
@@ -16,11 +35,51 @@ function TwoStageExaCore(args...; backend = nothing, concrete = Val(false), kwar
...
+) where {T, VT <: AbstractVector{T}, B, S <: TwoStageTags, C <: ExaCore{T, VT, B, S}}
+    ns = c.tags.ns
+    @assert ns > 0 "TwoStageExaCore must be created with ns > 0 to use EachScenario()"
+    nitr = length(gen.iter)
+    @assert nitr % ns == 0 "Number of constraints ($nitr) must be divisible by ns ($ns)"
+    nc_per_s = nitr ÷ ns
+    scenario_tags = [k for k in 1:ns for _ in 1:nc_per_s]
+    return constraint(c, gen; start, lcon, ucon, scenario = scenario_tags)
+end
+
```
sshin23: We should define the same syntax for `par`, `obj`, `expr`, `con!`

sshin23 (changes requested):

sshin23 (approved):

sshin23 (approved):

Issue #245, opened 2026-03-09
ExaModels AOT compilation


sshin23:
In order for ExaModels to be fully ahead-of-time (AOT) compilable without relying on the Julia runtime, several changes would be required. The main items are outlined below.

## Make ExaCore immutable and type-stable

Currently, `ExaCore` stores attributes with abstract types—most notably `core.obj` and `core.con`. This prevents full type inference and therefore blocks AOT compilation.

To resolve this, `ExaCore` should become an immutable struct whose type parameters encode the types of `core.obj` and `core.con`. As a consequence, `core` would need to be rebuilt whenever incremental changes are made.

This implies a change in syntax. For example:
```julia
    c = constraint(core, x[i] for i=1:10)   # old
    core, c = constraint(core, x[i] for i=1:10)   # new
```
If this change is adopted, the old syntax should be deprecated after one major release.

Optionally, we could introduce a higher-level syntax:
```julia
    core = ExaCore()

    core, x, y = variables(core, varspec(1:10), varspec(1:5))
    core, c1, c2 = constraints(core, conspec(x[i] for i=1:10), conspec(y[i] for i=1:5))
    core, o1, o2 = objectives(core, objspec(x[i]^2 for i=1:10), objspec(y[i]^2 for i=1:5))

    m = ExaModel(core)
```
Here, varspec, conspec, and objspec would simply cache their arguments
and internally call `variable(…)`, `constraint(…)`, and `objective(…)` when
executed within the corresponding `variables(…)`, `constraints(…)`, and
`objectives(…)` functions.

## Make Compressor type-stable

`Compressor` currently relies on `unique!` and certain for loops that introduce type instability. These would need to be rewritten to maintain full type stability.

## Relax some operator overloading specializations

Some operator overloads are currently overly specialized. For example, expressions such as
```julia
    ParSource() + 0
```
should avoid generating specialized methods. Relaxing these specializations would reduce compilation complexity and improve AOT compatibility.

## Proof-of-concept AOT modeling library

Once the above changes are implemented, the next step would be to develop a proof-of-concept AOT-compiled domain-specific modeling library built on ExaModels. This prototype should include:
-   a C interface
-   a standalone executable
It is not yet clear how difficult solver integration will be. As an initial target, Ipopt may be a simpler starting point than MadNLP.

cc: @hfytr

sshin23:
[...]

PR #252, opened 2026-03-11
AOT Compilation

hfytr:
solves #245 
@sshin23

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -154,41 +154,6 @@ Constraint Augmentation
 )
 
 
-mutable struct MutableExaCore
-    core::ExaCore
-end
-
-function variable(c::MutableExaCore, args...; kwargs...)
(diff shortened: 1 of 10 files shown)


src/nlp.jl
```diff
@@ -224,6 +223,61 @@ Base.@kwdef mutable struct ExaCore{T,VT<:AbstractVector{T}, B, S}
     tags::S = nothing
 end
 
+function mutate_core(
```
sshin23: A simpler way:
```julia
ExaCore(c::C; kwargs...) where C <: ExaCore = ExaCore(
    ;
    zip(fieldnames(C), ntuple(i -> getfield(c, i), Val(fieldcount(C))))...,
    kwargs...,
)
```

sshin23 (commented):
Looks great. We should add testing to ensure type stability

PR #253, opened 2026-03-16
Fix oneAPI and MacOS CI

michel2323:
Trying to make the CI more stable.

1. oneAPI was again hanging for whatever reason. I reset the Julia depot on the runner, but also added sequential precompiling.
2. The GH Mac OS runner was failing at times. I switched that runner also to aarch64. This should speed up the tests and avoid the complexity of Rosetta emulation. Maybe that helps.


sshin23:
Closing as CI issues have been addressed in #256

PR #255, opened 2026-04-02
AOT compilation

sshin23:
Succeeds #252


sshin23:
## Summary of changes relative to `origin/main`

### 1. Core API redesign — functional, immutable model building

**`ExaCore` is now an immutable struct** with full type parameters `{T,VT,B,S,V,P,O,C,R}`. The old mutable struct API (`variable()`, `parameter()`, `objective()`, `constraint()`, `constraint!()`, `subexpr()`) is replaced by a functional API where every `add_*` call returns `(new_core, handle)`:

```julia
c, x  = add_var(c, n; start=0, lvar=-Inf, uvar=Inf)
c, θ  = add_par(c, start_array)
c, o  = add_obj(c, expr for i in ...)
c, g  = add_con(c, expr for i in ...)
c, s  = add_expr(c, expr for i in ...)
c     = add_con!(c, g, contrib for i in ...)   # augment existing constraint
```

The macro layer (`@var`, `@par`, `@obj`, `@con`, `@con!`, `@expr`) wraps this and updates `core` in the calling scope.

Renamed types:
- `Subexpr` / `ReducedSubexpr` → `Expression` (always inline, no auxiliary variables)
- `ParameterSubexpr` → `Parameter`
- `Variable{S,O}` → `Variable{D,S,O}` (added disambiguation index `D` for variable block ordering)
- `Objective{R,F,I}` → `Objective{F,I}`, `Constraint{R,F,I,O}` → `Constraint{F,I,O}` (removed nested `inner` linked-list field)

Removed sentinel types: `VariableNull`, `ParameterNull`, `ObjectiveNull`, `ConstraintNull`.

Objectives and constraints are now stored as **tuples** instead of linked-list chains (dispatched via `Tuple{}` base case + `first`/`Base.tail` recursion). This gives fully concrete types at compile time — the key enabler for juliac.

---

### 2. AOT compilation via `juliac --trim=safe`

The main goal of this PR. Several changes were needed to make expression trees fully type-stable:

**`Const{T} <: AbstractNode`** (`src/specialization.jl`): a constant wrapper whose *value* lives in a field (not the type parameter), so `Const{Float64}` is always the same concrete type regardless of value. Unlike `Val{V}`, this does not trigger recompilation or produce abstract `Val{<:Real}` widening that blocks `--trim=safe`.

**Auto-const wrapping in macros** (`_maybe_const`, `_auto_const_gen`): free numeric symbols and literals inside `@obj`/`@con`/`@expr` generator bodies are automatically wrapped in `Const(...)`, preventing abstract `Val{<:Real}` when runtime values enter expression trees.

**`exa_sum` / `exa_prod`** with `SumNode{I}` / `ProdNode{I}` and `Val{N}`-based specialization for type-stable sum/product nodes over unit ranges. The `@obj`/`@con`/`@expr` macros rewrite `sum(body for j in range)` into `exa_sum(j -> body, Val(range))`, hoisting `Val(range)` outside the closure boundary in a `let` block — necessary because Julia's inference cannot propagate constants through closure boundaries.

**`NaNSource{T}`** replaces `nothing` for sparsity-detection passes, giving concrete types to the analysis and eliminating abstract dispatch paths.

**Val-based sparsity detection** in `src/simdfunction.jl`: `_gr_val` / `_hr0_val` dispatch trees compute gradient/Hessian NTuple sizes purely through types, replacing `@generated` functions and `unique(Any[])`/`Set{Any}` (which juliac cannot handle). Identity-based dedup via `_ident_unique` replaces `Set`.

**AOT apps and tests**: `test/LuksanVlcekApp.jl` and `test/COPSApp.jl` are standalone `@main` executables that compile with `juliac --output-exe a --trim=safe --experimental`. `test/NLPModelsIpoptLite.jl` is a minimal Ipopt C-API wrapper (via `ccall` to `libipopt`) designed for AOT use (no dynamic dispatch on the solver side, uses `@cfunction` with concrete model types). `test/JuliaCTest/JuliaCTest.jl` runs these compilations in CI.

---

### 3. Multi-generator `add_con`

A single `add_con` call can now accept multiple generators to build a constraint block where different generators contribute terms to the same set of constraint rows:

```julia
c, g = add_con(c,
    (x[i] for i in 1:N),                  # first contribution
    (i => sin(x[i]) for i in 1:N);        # second contribution (augmentation)
    lcon = 0.0, ucon = 0.0,
)
```

Internally, the first generator creates the constraint via the standard `add_con` path (allocating `N` constraint rows with the given bounds). Each subsequent generator is automatically forwarded to `add_con!`, which adds its terms to the same rows. The `i => expr` syntax in augmentation generators maps each expression to the appropriate constraint row via `Pair`-based indexing.

This is syntactic sugar for the explicit two-step pattern:
```julia
c, g = add_con(c, x[i] for i in 1:N; lcon=0.0, ucon=0.0)
c, _ = add_con!(c, g, i => sin(x[i]) for i in 1:N)
```

The multi-gen form is more concise and keeps all contributions to a constraint block visually grouped together. The keyword arguments (`lcon`, `ucon`, `start`, `name`) apply to the constraint block created by the first generator; subsequent generators inherit the block dimensions.

---

### 4. Subexpression changes (`@expr` / `add_expr`)

`subexpr(core, gen; reduced, parameter_only)` on `main` supported three modes:
1. **Lifted** (default, `reduced=false`): introduced auxiliary variables and equality constraints, returning a `Subexpr` handle indexable like a `Variable`.
2. **Reduced** (`reduced=true`): inlined the expression directly when indexed, returning a `ReducedSubexpr` — no auxiliary variables.
3. **Parameter-only** (`parameter_only=true`): evaluated once when parameters are set, cached values in θ, returning a `ParameterSubexpr`.

**This PR keeps only the reduced (inline) mode**, now exposed as `add_expr` / `@expr` → `Expression`:

- ✅ **Continued**: reduced/inline substitution — `s[i]` substitutes the expression tree directly into the enclosing objective or constraint. This is now the only behavior of `add_expr` / `@expr`. Multi-dimensional indexing is supported.
- ❌ **Discontinued**: lifted mode (auxiliary variables + equality constraints via `Subexpr`). This introduced abstract-typed constraint chains that blocked AOT compilation and added unnecessary NLP variables.
- ❌ **Discontinued**: parameter-only mode (`ParameterSubexpr`). The recomputation machinery (`_recompute_param_subexprs!`, `param_subexpr_fns`, etc.) is removed. Users who need parameter-dependent expressions can use `@expr` with `@par`-indexed values instead; recomputation happens naturally through the expression tree.

---

### 5. `ExaModelsLinearAlgebra` → `ExaModelsStaticArrays` (discontinued/replaced)

The `ExaModelsLinearAlgebra` extension on `main` (triggered by `using LinearAlgebra`) provided:
- Null-node pass-throughs for math functions (`inv`, `abs`, `sqrt`, etc.)
- Scalar Null arithmetic with zero/one elimination
- `zero`/`one`/`conj`/`adjoint`/`transpose` for `AbstractNode`

**This extension is discontinued.** Its functionality is split:
- The Null-node arithmetic and pass-throughs are now handled by the core library (built into `specialization.jl` and the `Const`/`Null` dispatch system).
- The `StaticArrays` integration (enabling `det`, `cross`, `norm`, `dot`, etc. on ExaModels expression nodes) is provided by a new **`ExaModelsStaticArrays`** extension, triggered by `using StaticArrays` (which pulls in `LinearAlgebra` as a co-trigger). This extension defines `zero`, `one`, `conj`, `adjoint`, `transpose`, `LinearAlgebra.dot`, and `StaticArrays.arithmetic_closure` for `AbstractNode`.

The `LinAlgTest` test module (1560 lines) covers AD correctness for all supported linear algebra operations.

---

### 6. Two-stage stochastic programming

The separate `TwoStageExaModel` wrapper (552-line `src/two_stage.jl`) is **deleted**. Two-stage support is now integrated into the core `ExaCore` via a tag system:

`TwoStageTags{VI}` (`src/tags.jl`) stores per-variable and per-constraint scenario indices. `TwoStageExaCore` is a type alias for `ExaCore{T,VT,B,S} where S <: TwoStageTags`. The `@var`/`@con` macros accept a `scenario=` keyword to assign scenario indices.

---

### 7. Variables constrained to a generator

```julia
c, x = add_var(c, f(p) for p in params)
```

Creates `length(params)` variables, each tied via an equality constraint to the corresponding generator value. Internally calls `add_var` for the block, then `add_con` to bind `x[j] == gen.f(orig)` for each element.

---

### 8. Bug fixes

- **`ExaCore(; backend=CUDABackend())` silently created a CPU model**: the no-T constructor did not forward `backend` to the inner call. Fixed.
- **GPU variable bounds**: uses `convert_array(b, backend)` instead of `adapt` which silently copied GPU arrays to CPU.
- **`add_var(c; kwargs...)` (scalar variable)**: now returns `Var(offset+1)` (an `AbstractNode`) so it's directly usable in expressions.
- **`obj_weight` type safety**: casts to `eltype(...)` to avoid Float64/Float32 mismatch on Metal and other Float32 backends.

---

### 9. Function registration overhaul

`src/functionlist.jl` and `ext/ExaModelsSpecialFunctions.jl` are regenerated by new dev-time scripts (`deps/generate_functionlist.jl`, `deps/generate_specialfunctions.jl`) using Symbolics.jl. All Float64 constant literals are replaced with type-generic helpers (`_clog2(x)`, `_cpi(x)`, etc.) for Float32 compatibility. New functions: `sinpi`, `cospi`, `sinc`, `expm1`, `atan(y,x)`, `hypot`, and more.

---

### 10. Backend / extension changes

- **Deleted**: `ExaModelsAMDGPU.jl`, `ExaModelsCUDA.jl`, `ExaModelsOneAPI.jl` — GPU array conversion now handled via `Adapt.jl` / `KernelAbstractions` only.
- **Simplified**: `ExaModelsOpenCL.jl` — only `sort!` override remains.
- **New**: `ExaModelsMetal.jl` — Float64→Float32 conversion, `default_T(::MetalBackend) = Float32`.
- **New**: `ExaModelsStaticArrays.jl` — replaces `ExaModelsLinearAlgebra` (see §5).

---

### 11. Dependency changes

| | Before | After |
|---|---|---|
| `Adapt` | indirect | direct dependency |
| `StaticArrays` | — | weak dep (new) |
| `CUDA`/`AMDGPU`/`oneAPI` | explicit extensions | via KernelAbstractions only |
| `ExaModelsLinearAlgebra` | extension | **removed** (→ `ExaModelsStaticArrays`) |

Package version: 0.9.3 → 0.9.7.

PR #256, opened 2026-04-04
AOT compilation

sshin23:
Succeeds https://github.com/exanauts/ExaModels.jl/pull/252 https://github.com/exanauts/ExaModels.jl/pull/255
https://claude.ai/code/session_01QsVaXnG1Cw7LdtCzvVYRoo


sshin23 (commented):
Good overall — the AOT infrastructure, benchmark suite, and CI improvements are solid. A few things to address:

## Must fix

**Binary benchmark results committed despite `.gitignore`**

`benchmark/benchmark-results-current.jld2` and `benchmark/benchmark-results-main.jld2` are binary files being tracked by git. The `benchmark/.gitignore` added in the same commit says `*.jld2`, but git won't ignore files that are already staged. These should be removed from the PR (`git rm --cached benchmark/benchmark-results-*.jld2`). Committing binary result files into the repo will grow it on every benchmark run.

**Comment/code mismatch in `_wrap_free_symbols`**

The function docstring says:

> Number literals and constant sub-expressions (those that reference no iterator variable) are wrapped in `Constant{T}()` so that algebraic simplification rules … fire at model-construction time.

But the code does the opposite:

```julia
if !(expr isa Expr)
    # Number literals, Symbols, LineNumberNodes — return as-is.
    return expr
end
```

Literals are passed through unchanged. Either the code needs to actually wrap literals in `Constant{V}()` (for the optimization to work from macros), or the docstring needs to be corrected to say literals are left as-is and handled by the `AbstractNode * Real` dispatch. As-is, the comment is actively misleading.

## Minor

**Auto-benchmark on every CI success is expensive**

The new `workflow_run` trigger runs ALL backends automatically whenever the "build" workflow succeeds on a PR. GPU runners (CUDA, AMDGPU, oneAPI, Metal) are likely self-hosted and constrained — running the full benchmark suite on every push may queue-starve or incur unexpected costs. Consider defaulting to `nothing`-only on auto-trigger and keeping the full suite opt-in via comment.

**`ExaModelsLinearAlgebra` is silently dropped**

`ExaModelsLinearAlgebra` is removed from `Project.toml` and `LinAlgTest` is commented out in `runtests.jl`. Is the extension intentionally deleted? The ext file `ext/ExaModelsLinearAlgebra.jl` appears in the changed files list. If it's being removed it should be explicit in the PR description.

**`_needs_overload` fix is correct and good** — the `!hasmethod` approach was indeed broken for Base generics like `max`.

**`c.con` → `c.cons` in `ExaModelsKernelAbstractions.jl`** — correct bug fix, worth calling out explicitly in the PR body.

sshin23:
Hi @frapac @amontoison @pulsipher @jbcaillau @apozharski @Robbybp @klamike — [PR #256](https://github.com/exanauts/ExaModels.jl/pull/256) This is a large diff and I wouldn't ask for a full review, but it would be great if you could take a quick look before I merge. Perhaps the biggest change is the modeling syntax. (I know that this will be painful for those who've been developing things based on ExaModels, but if this has to be done, it is much better to do it now than later.) Here's the context. 

---

## Goal

Enable ahead-of-time (AOT) compilation via `juliac --trim=safe` without fundamentally changing what ExaModels *does*.

---

## Why this required a redesign of `ExaCore`

The previous `ExaCore` stored objectives, constraints, etc. as fields of abstract or union types. `juliac --trim` requires every type in the call graph to be concrete and statically resolvable, and abstract field types make that impossible.

The fix: `ExaCore{T,VT,B,S,V,P,O,C,R}` now carries the full type of every component (`V` for variables, `O` for objectives, `C` for constraints, …) as a type parameter. As a result, **`ExaCore` is recreated as an immutable value every time a variable, constraint, or objective is added** — rather than mutating a single object. This is the main structural change.

---

## New syntax — macros hide the immutability

Rebuilding an immutable struct on every `add_var` / `add_con` is unfamiliar to users coming from JuMP. The macro interface hides this: each macro rebinds `core` in the calling scope automatically.

```julia
c = ExaCore()

@var(c, x, N)               # c is rebound; x is in scope as a Variable
@var(c, y, 1:M; lvar = -1)  # y in scope too
@par(c, θ, data)             # θ is a Parameter vector

@obj(c, (x[i] - θ[i])^2 for i in 1:N)

@con(c, g, x[i] + x[i+1] for i in 1:N-1; lcon = 0, ucon = 1)  # g in scope

model = ExaModel(c)
```

Named variables/constraints are also registered on `c` for later access as `model.x`, `model.g`, etc. The macro names are chosen in a way that doesn't clash with JuMP.

Note that 
```julia
@var(c, x, N)
@con(c, g, x[i] + x[i+1] for i=1:N-1)
```
is equivalent to 
```julia
c, x = add_var(c, N; name = Val(:x))
c, g = add_con(c, x[i] + x[i+1] for i=1:N-1; name = Val(:g))
```
---

## Subexpression changes

- **Parameter-only subexpressions are dropped.** There was no real use case for a subexpression that only referenced parameters and not variables; removing it simplified the type hierarchy.
- **Reduced formulation is now the default.** Previously users had to opt in; now `@expr` always produces a reduced (shared) subexpression.
- **Lifted formulation via `@var`.** If you want an explicit auxiliary variable (the old "lifted" formulation), just introduce it with `@var` and add an equality constraint with `@con`. This keeps the two formulations orthogonal and makes the lifted case explicit:

---

## `ExaModelsLinearAlgebra` is removed

ExaModels is designed around scalar kernel formulations. The `LinearAlgebra` extension was pushing against that by letting users write matrix/vector AD — but the compilation cost scales with `N` (it blows up for large matrices), and it risks leading users toward a usage pattern that ExaModels is not meant to support. If vector AD is needed, it should be handled by an external AD tool. Removing it now while the user base is small.

---

## Benchmark script

A `benchmark/` directory is now included. The CI benchmark workflow runs automatically on each PR (CPU backend) and can be triggered for GPU backends via a comment. This lets us catch performance regressions before merging.

---

Happy to answer questions or discuss any of the design choices. Thanks for taking a look!

amontoison:
I am against removing `ExaModelsLinearAlgebraExt` if no alternative is provided.

Jean-Baptiste (@jbcaillau) traveled to Chicago such that we work on that with me and @michel2323.

We also spent some funding to develop it.

sshin23:
@amontoison We can discuss alternative implementation plans. Vector functions can be supported: e.g., https://github.com/exanauts/ExaModels.jl/pull/230#issuecomment-4158492065, https://github.com/exanauts/ExaModels.jl/issues/258, but supporting general sparse matrix multiplication is likely not a good idea

amontoison:
I don't think that we support sparse matrix multiplications.
The extension is for `LinearAlgebra.jl` and not `SparseArrays.jl`.

I am happy to discuss about it.
But I really don't want to break support for Jean-Baptiste, by respect for his work and the collaboration under the funding FACCTS.

Michel will need to do a report about what was achieved with the funding FACCTS (UChicago -- INRIA) and removing the extension has on impact on that too.

sshin23:
run benchmark

src/nlp.jl
```diff
@@ -138,8 +126,17 @@ Constraint
...
+    ConstraintAug
+
+An augmentation layer added to an existing [`Constraint`](@ref) via
+[`add_con!`](@ref) / [`@con!`](@ref). Each element of the iterator yields an
+`idx => expr` pair: `expr` is accumulated into the constraint row identified
+by `idx` at evaluation time. Multiple `ConstraintAug` objects can be stacked on
+the same base constraint to aggregate contributions from several data sources
+(e.g. summing arc flows into nodal balance constraints).
+"""
+struct ConstraintAug{F,I,D} <: AbstractConstraint
```
frapac: maybe rename as `ConstraintAugmentation` to be more explicit?

src/nlp.jl
```diff
@@ -500,21 +544,26 @@ julia> using ExaModels
...
+
+julia> y
 Variable
 
   x ∈ R^{9 × 3}
 
 ```
 """
-function variable(
+function add_var(
```
frapac, line 567: I would maybe rename this function as `add_variable`? But I understand this may break the connection with the macro `@var`, whose name is constrained as we don't want to clash with JuMP.

src/nlp.jl
```diff
@@ -524,43 +573,144 @@ function variable(
...
+julia> c, θ = add_par(c, ones(10));
+
+julia> θ
 Parameter
 
   θ ∈ R^{10}
 ```
 """
-function parameter(c::C, start::AbstractArray;) where {T,C<:ExaCore{T}}
+function add_par(c::C, start::AbstractArray; name = nothing) where {T,C<:ExaCore{T}}
```
frapac, line 664: Same here, I would maybe rename this function `add_parameter`

src/nlp.jl
```diff
@@ -716,18 +887,61 @@ function constraint(
...
 
-Adds constraints specified by a `expr` and `pars` to `core`, and returns an `Constraint` object.
+When `name` is given as `Val(:name)`, the constraint is also accessible as
+`core.name` or `model.name`.
+
+Prefer the generator form (`add_con(core, gen)`) for typical use; this form
+is intended for code that builds expression trees programmatically.
 """
-function constraint(
+function add_con(
```
frapac, line 899: Maybe rename as `add_constraint`?

src/nlp.jl
```diff
@@ -848,201 +1130,68 @@ _infer_subexpr_dims(itr::Base.Iterators.ProductIterator) = itr.iterators
 _infer_subexpr_dims(itr) = (length(collect(itr)),)  # fallback
 
 """
-    subexpr(core, generator; reduced=false, parameter_only=false, dims=nothing)
+    add_expr(core, generator; name = nothing)
```
frapac, line 1074: Maybe rename as `add_expression`?

frapac (commented):
This review looks mostly at the functions exposed in the API, and suggests some renaming to get more explicit naming in the API:
- `add_var` -> `add_variable`
- `add_con` -> `add_constraint` 
- `add_par` -> `add_parameter` 
Let me know if that makes sense for you. 

Your PR summary is very appropriate to understand the new changes in ExaModels. Overall, the direction looks good to me, and I think it's worth the breaking change. 

I just have one quick question regarding the performance. If we have to re-create a new `ExaCore` each time we add a variable/constraint, do we observe an impact in the performance when creating large model (like SCOPF)? And how much pressure does this put to the GC?

michel2323:
I have to agree with @amontoison here. At this point, ExaModels.jl has some users. Below is what breaks with this PR. In summary, that would be all of @jbcaillau's codes, my codes (including MathOptAI.jl integration), and potentially more users out there. Work that has been done over the last couple of months, and nobody has objected. Not to mention the TwoStageModel is on our HPC4EI deliverables list.

I see the benefit for AOT, however, some of these changes may only be needed because of current limitations in juliac
  --trim=safe on Julia 1.12, and might not be necessary as the trimmer matures. Also, not supporting LTS is just another hint to me that it is a bit fast-tracked. Moreover, I'd argue that there are not many users who actually use AOT now. I'd also argue that most users are interactive users who don't do AOT.

My suggestion is to support both pathways in another release and then discuss a way forward. Merging this as is and releasing leaves no choice but to fork the package.

### 1. Mutable, fluent model building

  The single biggest behavioral break. On `main`:

  ```julia
  c = ExaCore()
  x = variable(c, 10)
  constraint(c, x[i] + x[i+1] for i in 1:9)
  objective(c, x[i]^2 for i in 1:10)
```
 After this PR, ExaCore is immutable. Every builder call returns a new core, so the user must thread it through:
```julia
  c = ExaCore()
  c, x = add_var(c, 10)
  c, _ = add_con(c, x[i] + x[i+1] for i in 1:9)
  c, _ = add_obj(c, x[i]^2 for i in 1:10)
```
  or use the new @var / @con / @obj macros (which rebind c behind the scenes). Either way, every existing
  ExaModels script, tutorial, and downstream package breaks at parse time with no deprecation warning.

  ### 2. Matrix/vector algebra in expressions

  On main, with LinearAlgebra loaded, you could write natural linear-algebra expressions in objectives and
  constraints:
```julia
  constraint(c, A * x - b)        # vector constraint, expands row-wise
  objective(c, dot(c, x))         # linear objective
  constraint(c, norm(x - x0)^2 - r^2)
  objective(c, tr(M' * M))
```
  After this PR these all throw MethodError. Users must hand-expand every matrix/vector op into scalar generators
  (add_con(c, sum(A[i,j]*x[j] for j in 1:n) - b[i] for i in 1:m)). For anyone modeling problems with matrix data
  this is a real ergonomic regression — and it's the kind of thing JuMP users specifically come to ExaModels for.

  ### 3. Lifted subexpressions

  On main, the default subexpr(c, gen) form created auxiliary variables s plus defining equality constraints s[i]
  == expr(i), so an expensive subexpression evaluated once and was reused across constraints/objectives without
  re-differentiating through it. This is a standard NLP modeling technique. After this PR only the "reduced" (pure
   substitution) form survives — the same expression in 10 places gets compiled and differentiated 10 times.

  ### 4. Parameter-only subexpressions

  subexpr(c, gen; parameter_only=true) cached values that depended only on θ, recomputed them automatically in
  set_parameter!, and skipped them entirely during the optimization hot path. For parametric workflows (MPC,
  scenario sweeps, sensitivity studies) where the same model is solved repeatedly with different parameter values,
   this was a meaningful speedup. The whole machinery — _recompute_param_subexprs!, the param_subexpr_* fields on
  ExaCore, the ParameterSubexpr type — is gone.

  ### 5. Two-stage stochastic models

  TwoStageExaModel(nd, nv, ns, θ_sets) do c, d, v, θ, ns, nv, nθ ... end and the entire block-decomposition API
  around it (recourse_var_indices, extract_recourse_vars!, total_vars, cons_block_indices, grad_recourse_indices,
  etc.) are removed. The replacement TwoStageExaCore only attaches scenario tags to variables and constraints —
  there is no longer a way to extract per-scenario blocks of x, g, or the gradient. Anyone using ExaModels for
  two-stage stochastic programming loses their core abstraction.

  ### 6. Interactive REPL responsiveness

  Not a deleted feature, but a fundamental behavioral change: because ExaCore now encodes the tuple-types of all
  variables/parameters/objectives/constraints in its type parameters, every add_* call triggers a fresh JIT
  specialization. A REPL session that builds a 20-block model pays compile latency on every single add_con call,
  with the cost growing as the tuple gets longer. On main, all the per-shape compilation happened exactly once at
  ExaModel(c) time. The new model is great for AOT (juliac --trim=safe) and bad for notebooks / Pluto / scripting.

  ---
  TL;DR for users: if you currently (a) build models incrementally at the REPL, (b) use A*x or dot / norm / tr in
  constraints, (c) rely on lifted or parameter-only subexpr, or (d) use TwoStageExaModel, this PR breaks your
  workflow with no deprecation path.

pulsipher:
My principal concern with this change would pertain to performance of incremental builds. With InfiniteExaModels, AOT is not of much concern and I fear these changes would introduce appreciable performance degradations, though I would need to make an experimental PR to double check. I also would hesitate to not support the LTS version of julia.

sshin23:
@frapac @pulsipher great point. Model creation performance actually improves with this. Note that we don't recreate the array (arrays are reused from the previous build of ExaCore). What's recreated is just the references to arrays, etc, which is small. A greater benefit comes from type stability. If we compare model creation performance, it's actually much faster with this PR.

`main`:
```julia-repl
julia> using ExaModels, BenchmarkTools

julia> function exa_rosenrock_model(backend, N)
           c = ExaCore(backend = backend)
           x = variable(c, N; start = (mod(i, 2) == 1 ? -1.2 : 1.0 for i = 1:N))
           constraint(c, 3x[i+1]^3 + 2*x[i+2] - 5 + sin(x[i+1] - x[i+2])sin(x[i+1] + x[i+2]) + 4x[i+1] - x[i]exp(x[i] - x[i+1]) - 3 for i = 1:(N-2))
           objective(c, 100*(x[i-1]^2 - x[i])^2 + (x[i-1] - 1)^2 for i = 2:N)
           return ExaModel(c)
       end
exa_rosenrock_model (generic function with 1 method)

julia> @btime exa_rosenrock_model(nothing, 100);
  22.526 μs (317 allocations: 24.13 KiB)
```

This PR:
```julia-repl
julia> using ExaModels, BenchmarkTools

julia> function exa_rosenrock_model(backend, N)
           c = ExaCore(; backend)
           @var(c, x, N; start = (mod(i, 2) == 1 ? -1.2 : 1.0 for i = 1:N))
           @con(c, s, 3x[i+1]^3 + 2 * x[i+2] - 5 + sin(x[i+1] - x[i+2])sin(x[i+1] + x[i+2]) + 4x[i+1] - x[i]exp(x[i] - x[i+1]) - 3 for i = 1:(N-2))
           @obj(c, 100 * (x[i-1]^2 - x[i])^2 + (x[i-1] - 1)^2 for i = 2:N)
           return ExaModel(c)
       end
exa_rosenrock_model (generic function with 1 method)

julia> @btime exa_rosenrock_model(nothing, 100);
  12.972 μs (173 allocations: 17.40 KiB)
```

sshin23:
@pulsipher @michel2323 the intention for dropping LTS from CI was to make CI lighter, but LTS should continue to work except for the new AOT compatibility. I will add LTS back

sshin23:
@pulsipher @michel2323 — I agree that most users will rely on ExaModels in a JIT-compiled setting.

That said, I think you both recognize the value of AOT compilation. Enabling AOT compatibility would significantly broaden what users can do with ExaModels. At the moment, the reliance on the Julia runtime is a non-starter for many applications; being able to generate binaries and expose a C interface would open up many opportunities for model developers.

If we want AOT compatibility, this kind of syntax redesign is likely unavoidable. The core issue is that maintaining model cache within ExaCore in a mutable form makes the modeling script inherently type-unstable, which fundamentally limits AOT support. ExaModels is actually well positioned for AOT compilation, since much of ExaCore’s internals are already explicitly typed.

Overall, I understand that this change would be breaking. But if we believe this direction is necessary—and I do—then it’s probably better to make the transition now, before more packages build dependencies on the current design.

pulsipher:
Sounds good. I misread the changes to the project.toml on the julia versioning. I don't have any major concerns if there are no performance degradations, but I would like to make a draft PR to InfiniteExaModels to properly benchmark our use cases to make sure there are no gotchas before merging this.

sshin23:
@michel2323 Regarding deprecation, seeing the responses, I think it is a good idea. I'll add deprecation.

sshin23:
@michel2323 I liked your work on subexpr, and I kept most of it. Lifted subexpression is continued, but it is slightly differently labeled. We simply call it `@var`, since we introduce a new variable, and constriant is automatically created together. E.g., we used to do
```julia
subexpr(core, sin(x[i]) for i=1:10; reduced = false)
```
but now we do
```julia
@var(core, sin(x[i]) for i=1:10)
```
I believe this is quite intuitive. 

The only thing that was discontinued is the parameter-only subexpr. My rationale was that this is something a modeler can easily implement on their own, and there is not much performance loss even if we need to recompute them. A more important performance issue is likely the recomputing of the generic expression.

sshin23:
@michel2323 regarindg two-stage.

The previous two-stage stochastic model was removed in https://github.com/exanauts/ExaModels.jl/pull/226, and this only removes the dead test and makes it algined with the current tagging-based approach. I think we decided to move forward with the following syntax:
```julia
TwoStageExaCore(nscen = 10)
variable(core, 1:4, EachScenario())
# the same as variable(core, 1:4, 1:10; scenario = (j for i=1:4, j=1:10))
```

sshin23:
@michel2323, regarding interactive REPL responsiveness, I agree this is a legitimate long-term concern. Compilation will likely become more expensive as the entire ExaCore becomes more explicitly typed.

That said, I’d argue—and it may be worth stating this explicitly—that the intended users of ExaModels are heavy users, particularly those developing domain-specific models (i.e., it’s not necessarily a great fit for an optimization class setting). Our priority should be delivering optimized runtime performance: the five AD callbacks (obj, cons!, grad!, jac_coord!, hess_coord!) are first-class, while model construction time is second-order, and compilation time is a lower priority unless it scales poorly with problem size. In this case, the change increases compilation cost, but likely only by a constant factor, without introducing adverse scaling with problem dimension.

As demonstrated in https://github.com/exanauts/ExaModels.jl/pull/256#issuecomment-4199319099, we improve runtime model construction time, potentially at the expense of higher compilation time.

sshin23:
@michel2323, regarding matrix multiplication, I’d go back to the point that the intended users of ExaModels are primarily heavy users. Matrix multiplication is an area where ExaModels will likely never excel, so officially supporting it doesn’t make much sense to me. I’d prefer that users lean into what ExaModels is designed for, rather than spending effort forcing use cases it isn’t meant to handle. For domain-specific model developers, performance should take precedence over convenience, and writing constraints in scalar form doesn’t seem overly burdensome.

The Linear Algebra extension PR was merged before I had a chance to review it closely—thanks for the patience on my slow response—and I believe I raised these concerns in our initial meeting.

More broadly, I do think supporting vector constraints is a good idea, but I don’t see a strong case for matrix multiplication. I asked for concrete use cases here: https://github.com/exanauts/ExaModels.jl/pull/230#issuecomment-4158434346, since we may be able to design something more appropriate for optimal control applications without relying on generic matrix operations. This is something we can discuss further in the meeting with @jbcaillau.

Regarding the MathOptAI extension, I think we can support it. It’s under 1000 lines of code and doesn’t seem to depend heavily on ExaModels internals. Updating the syntax should be relatively straightforward, and I’m happy to help with that if needed.

pulsipher:
As a follow up, I drafted the PR for InfiniteExaModels and did some benchmarking (see https://github.com/infiniteopt/InfiniteExaModels.jl/pull/31). I don't find any significant increase or decrease in performance, so I have not objections to this PR. 

I also agree with @frapac that using full names for the functions/macros seems like a better choice (e.g., `add_variable` instead of `add_var`).

klamike:
Very excited for this, great work! I'll try to find some time to review it more closely, maybe even try using it from Python together with libMad...

For now, one comment is that it would be good to have some documentation page to explain how to update existing codes from ExaModels 0.9 to ExaModels 0.10. It can be verified by asking your favorite coding agent to follow the guide (without letting it read the source code..) to update e.g. ExaModelsPower or the MathOptAI extension, making note of the mistakes it makes along the way.

test/LuksanVlcekApp.jl/a
klamike, line 1: is this file used?

src/nlp.jl
```diff
@@ -3,19 +3,24 @@ abstract type AbstractParameter end
...
+"""
+    Variable
 
-struct Variable{S,O} <: AbstractVariable
+A handle to a block of optimization variables added to an [`ExaCore`](@ref) via
+[`add_var`](@ref) / [`@var`](@ref). Use indexing (e.g. `x[i]`) to reference
+individual entries in objective and constraint expressions. Retrieve solution
+values with [`solution`](@ref).
+"""
+struct Variable{D,S,O} <: AbstractVariable
```
klamike: If I understand correctly the `D` is to make each block of variables have its own type? But we never dispatch on that it seems. And Parameter doesn't have it. Is this some AOT magic? Locally the tests pass (cpu and Metal) without it..

src/nlp.jl
```diff
@@ -200,44 +199,105 @@ An ExaCore
...
+    ;
+    zip(fieldnames(C), ntuple(i -> getfield(c, i), Val(fieldcount(C))))...,
+    kwargs...,
+)
+@inline default_T(backend) = Float64
+@inline append_var_tags(::Nothing, backend, len) = nothing
+@inline append_con_tags(::Nothing, backend, len) = nothing
+
+
+@inline depth(a::Tuple) = length(a)
```
klamike: is this needed? why not just use length directly?

test/NLPModelsIpoptLite.jl/src/NLPModelsIpoptLite.jl
klamike, line 1: it might be good to make this a separate package, even if unregistered

sshin23: This is more of a temporary package, and I'm planning to make PR to NLPModelsIpopt and remove this package in the future

sshin23:
@frapac @klamike if we do `add_variable` and `add_constraint`, the symbol name will clash with JuMP, and I don't want to cause too much pain with this. Any suggestions for the naming?

amontoison:
I like a small prefix like `exa_` for the symbols.
It is what I did with Uno (`uno_`) to avoid conflits with other libraries.

sshin23:
@amontoison `exa_add_variable` would be too wordy; `exa_variable` misses the verb add; I think the verb add makes it intuitive that we're not just creating a variable but adding the variable information to the core. I'll stick to `add_var`.

PR #215, opened 2026-02-05
Expression Support

hfytr:
This PR changes expressions to no longer re-evaluate on every reference, and instead pre-computes once for every call to jac_coord / hess_coord / etc.

PR #274, opened 2026-04-18
Expression Support (PR #215 rebased onto main)

sshin23:
## Summary

- Squash-merges PR #215 (Expression Support by @hfytr) onto current `main`
- Adds pre-computed subexpression caching: expressions are evaluated once per `obj`/`grad!`/`jac_coord!`/`hess_coord!` call instead of re-evaluated on every reference
- Resolves conflicts with two-stage model, getter/setter API, and other recent changes

## Changes from PR #215
- `src/graph.jl`: new `Expression`/`AbstractNode` subtype for subexpressions
- `src/nlp.jl`: `ExaCore`/`ExaModel` carry expression storage
- `src/hessian.jl`, `src/jacobian.jl`, `src/gradient.jl`: expression eval threading
- `src/simdfunction.jl`: SIMD kernel updates


sshin23:
Closing — pushing changes directly to supersede PR #215 instead.

PR #283, opened 2026-06-05
Unpin GPUCompiler

michel2323:


KSepetanc:
Should also remove from weakdeps? No extension uses it and without compat entry I think it can be removed from there too.

PR #304, opened 2026-08-11
CI: coverage uploads from the self-hosted legs; OIDC everywhere

sshin23:
Only the hosted `github` job uploaded coverage, so the report never saw lines that only the self-hosted legs exercise: the KernelAbstractions/OpenCL extension paths (cpu leg), the CUDA device paths (gpu leg), and `ExaModelsC/src` entirely (its tests run nowhere else). This is a large part of why the repo reads 81%.

- One instrumented leg per distinct population: `cpu` on julia 1 (KA + PoCL), `gpu` on julia 1 + cuda, `examodelsc` on linux. The other matrix legs run uninstrumented as before.
- `julia-processcoverage` explicitly covers `src,ext` (and `ExaModelsC/src` for its job).
- All uploads authenticate over OIDC (`id-token: write`), matching CNLPModels.jl/cnlpmodels-py/MadNLP.jl, and the first upload from this PR also populates codecov's `madsuite-org/ExaModels.jl` slug.
- The `cnlpmodels-py` checkout in the examodelsc job points at madsuite-org.

A separate follow-up will add tests for whatever remains genuinely uncovered once this lands and the merged report shows the real number (a local full-matrix instrumented run is in progress to get that list ahead of time).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,7 +61,7 @@ jobs:
         julia-version: ['1', 'lts']
         julia-arch: [x64]
         os: [ubuntu-latest]
-    runs-on: ['self-hosted', 'examodels']
+    runs-on: ['self-hosted', 'shin-compute']
     steps:
     - name: Checkout code
       uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


sshin23:
Added: runner selection drops the `examodels` label — Linux legs select `['self-hosted', 'shin-compute']` (+ backend labels), mac legs `['self-hosted', 'macOS']` / `['self-hosted', 'gpu', 'metal']`, matching the rebuilt org-level fleet.

PR #307, opened 2026-08-12
Add ExaModels.SIMDMode: ExaModels as an MOI.Nonlinear AD backend

blegat:
Instead of ExaModels pretending to be a solver via ExaModels.Optimizer, any MOI solver that supports `MOI.AutomaticDifferentiationBackend` can now evaluate its nonlinear model through ExaModels by setting the backend to ExaModels.SIMDMode(; device). Written by Claude Fable 5, still need to check. Needs

- [ ] https://github.com/jump-dev/MathOptInterface.jl/pull/3046


odow:
This would still require a CPU <-> GPU switch at each iteration. Part of the reason for ExaModels.Optimizer is that the entire problem gets copied to the GPU, instead of just evaluating the derivatives.

ext/ExaModelsMOI.jl
```diff
@@ -903,4 +903,233 @@ function MOI.set(model::Optimizer, ::MOI.NLPBlock, nlp_data::MOI.NLPBlockData)
...
+Unlike `MOI.Nonlinear.Model`, this model accepts `MOI.ScalarAffineFunction`
+and `MOI.ScalarQuadraticFunction` objectives and constraints natively, so it
+must not be wrapped in `MOI.Nonlinear.ModelWithQuad`: ExaModels exploits the
+structure of affine and quadratic functions itself.
+"""
+mutable struct SIMDNonlinearModel
+    objective::Any
+    constraints::Vector{Tuple{Any,Any}}
+    SIMDNonlinearModel() = new(nothing, Tuple{Any,Any}[])
+end
```
odow, line 929: Why does this PR need a new model object? Can't we just build the ExaModel from the existing MOI.Nonlinear.Model?

blegat: Good question, the issue is that it is easier to do the conversion at once but `MOI.Nonlinear.add_constraint` gives things one by one. For example, we need `ordered_variables` which is given only at `MOI.Nonlinear.Evaluator`. This is kind of the `supports_incremental_interface` we have in MOI. So it means ExaModels's AD backend should receive the model as a whole. I'm starting to think the interface should simply be
```julia
ad_model = MOI.Nonlinear.evaluator(ad_backend, model, ordered_variables, features)
# Now we can call the callbacks. `ad_backend` can do
# whathever he wants as long as the callbacks work
```
Then so here ExaModels can implement `copy_to` so it does not need to create any new custom objects.
If often happens that we only want to give part of the model to `ad_model` but that's very easy to do with `MOI.Utilities.Filter`, CCOpt is doing this for complementarity constraints.

Incremental interface is nice for solver supporting modifications but it's annoying in other cases. We've already simplified our life greatly by allowing solvers to just implement `MOI.copy_to`, I think we just need to do the same here.

blegat: But if the AD model does not support incremental mode, it also means that any solvers needing an AD model cannot support it so I'll first check if there is no way to just work around it

blegat: Actually, your PR https://github.com/madsuite-org/ExaModels.jl/pull/290 makes it easy to support incremental interface, I'll just assume `ordered_variables` is trivial for now

PR #308, opened 2026-08-13
ExaModelsC: emit several models into one library

sshin23:
`compile_library` now takes any number of `:name => core` / `:name => (core, args...)` pairs and compiles them into one shared library, each model under its own symbol prefix — and, per model, compiles **any argument shape `ExaModel` takes**, emitting the matching instantiation surface:

```julia
compile_library("@grid",
    :acopf  => (ac_core, 100),                 # one integer → acopf_new(n)
    :struct => (s_core, 4,                     # several values → the builder ABI
                (v0 = fill(0.5, 4), lo = fill(-10.0, 4)),
                [(i = 1, w = 2.0, s = 1.0)]),
    :opf    => (opf_core, ac_opf_args, "case14.m"),  # argument function → opf_new_str
    :fixed  => small_core)                     # no instantiation data
```

**Multi-model:** the generated module suffixes all per-model state (`CORE_$p`, `MODELS_$p`, ...), so models keep separate cores, instance tables, and ids; the library FILE is named after `out`, the model names supply the prefixes.

**Structured instantiation:** recipes whose examples are not a single integer emit the ABI v2 schema + builder surface both consumers already implement — `P_schema` publishes the flattened field list, typed setters take values by name, `P_new_from_data` reassembles the `ExaModel` arguments. Builder storage is generated concretely from the example types (`Int64`/`Float64` exactly; anything looser is refused with the reason).

**Argument functions:** `argfun = f` (or a `Function` in the pair's second position — nothing callable can be a model argument, so the spelling is unambiguous) carries `f` into the library and calls it at run time: only `f`'s own argument crosses the boundary — one string (`P_new_str`) or one integer (`P_new`) — and the work that turns it into instantiation data happens inside. `f` must be a **named function owned by a package proper**: functions from scripts/REPL (unreachable from the generated app) and from package *extensions* (their module is not a name the app imports) are refused in the first second with the constraint named, rather than dying minutes into juliac.

**Every model exports `P_argkind() -> 0|1|2|3`** (fixed, `P_new(n)`, `P_new_str`, builder), so a consumer routes on declared shape instead of probing symbols; kind 1 covers n-is-the-size and n-goes-to-the-function indistinguishably **by design**. `P_nargs` (how many) and `P_argkind` (what shape) are the two halves a consumer handed only a library path needs.

**Subexpressions work in recipes:** `@add_expr` now stores its body as a node (substituted via a static `_reindex` walk) rather than a closure — a closure captures the `Variable`s it mentions, putting a placeholder's `ArgSource` into the closure's *type*, where `instantiate` cannot reach it. Verified against a measured main baseline (1617/0/2, identical), plus empirical probes of `sum(...)` inside bodies, including index-dependent summands (exact).

**Developed packages are pinned:** every package the caller tracks by path becomes a `[deps]` + `[sources]` entry of the generated app (not an import — resolution needs the pin; only the cores' own types need imports), so the app compiles the code the caller runs rather than the registry copy. Also in ExaModels proper: `instantiate`'s varargs are specialization-forced (`Vararg{Any,N}`) *and* the container walk is unrolled — both are needed; large cores (AC OPF scale) leave `map` closures unresolved under `--trim=safe` with the annotation alone.

Suite: **210/210** on Linux against current consumer masters — a three-model round-trip, the four-surface library (builder + one-knob + argfun:int + argfun:str in one file), Ipopt solves through both surfaces, refusal tests (strings, loose types, colliding names, script-, Main-, and extension-owned functions). Scale datum from independent review: an **18-model library assembles in 160 s with 0 verifier errors**, exact readback at an unseen size. GPU datum from downstream validation: a real modelling library's CUDA test slice (ExaModelsPower#57, 322 tests) runs green against this branch's instantiation path — the leg that was dead before the `collect` fix.

Note for path-dev consumers: this adds `Pkg` to ExaModelsC's dependencies — a stale manifest that path-devs ExaModelsC reports `Package ExaModelsC does not have Pkg in its dependencies` until `Pkg.resolve()`; that is the manifest, not the branch.

Consumer-side name selection (`CNLPModel(lib, :acopf, ...)`) is already on CNLPModels master; the `_new_str`/`_argkind` routing follows as separate consumer work.

### DIFF
--- a/ExaModelsC/Project.toml
+++ b/ExaModelsC/Project.toml
@@ -6,6 +6,7 @@ authors = ["Sungho Shin <sushin@mit.edu>"]
 [deps]
 ExaModels = "1037b233-b668-4ce9-9b63-f9f681f55dd2"
 JuliaC = "acedd4c2-ced6-4a15-accc-2607eb759ba2"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
 TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
--- a/ExaModelsC/src/ExaModelsC.jl
+++ b/ExaModelsC/src/ExaModelsC.jl
@@ -52,6 +52,25 @@ m = cnlpmodels.CModel(lib, 1000, prefix="rosen")
 Both consumers default `prefix` to `"rec"` when handed a library handle, while
 `compile_library` defaults it to the output directory's name — pass it
 explicitly unless those coincide.  The test suite exercises both.
+
+## Several models in one library
+
+A library can carry more than one model, each under its own symbol prefix.
+Name them, and the consumers select by the same name:
+
+```julia
+compile_library("@grid", :acopf => (ac_core, 100), :dcopf => (dc_core, 100))
+```
+```julia
+CNLPModel("@grid", :acopf, 100)                  # Julia
+```
+```python
+cnlpmodels.CModel("grid", 100, prefix="acopf")   # Python — a bare name, no sigil
+```
+
+They share the library file and, in a bundle, its one privatized ~80 MB Julia
+runtime — which is the reason to co-package a family of models rather than emit
+a library each.
 """
 module ExaModelsC
 
@@ -67,6 +86,15 @@ export compile_library
 # its own environment, never registered.  A constant UUID is therefore safe and
 # saves a UUIDs dependency.
 const _GEN_UUID = "b41c7e02-9f3d-4a58-8e6c-2d0f5a7c9b13"
+
+# One model's compiled form: its prefix, its concretized core, the example value
+# whose types `juliac` needs, and how `<prefix>_new(n)` rebuilds it (`field`).
+struct ModelSpec
+    prefix::String
+    core::Any
+    arg::Any
+    field::Any
+end
 
 """
     compile_library(out, core, args...; prefix = basename(out), trim = "safe",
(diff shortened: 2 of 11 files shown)


sshin23:
Pushed one commit: `@add_expr` now works inside a recipe.

**The problem.** `@add_expr` stored its body as the generator's closure and
re-invoked it at each point of use. Inside a model-building *function* that
closure captures the `Variable`s it mentions, so when a size is a placeholder
the `ArgSource` lands in the closure's type — where `instantiate` cannot reach
it, and where it cannot be rebuilt field-wise either, since Julia emits no
constructor for a closure type. ExaModels then refuses the core with
"instantiating this core left argument placeholders in it". (At top level the
same model appears to work — a global is captured by reference — which makes
the failure look intermittent when it is really about scope.)

**The change.** The body is stored as a node, built once against a
`DataSource`, and indexing an `Expression` substitutes into it (`_reindex`).
Nodes have constructors and dispatch, the operation lives in the type
parameter, so the walk is static: no reflection, no `@generated`. Once the
source is replaced by a concrete element — an integer, or the tuple a
multi-dimensional generator destructures — the lookups above it are performed
rather than rebuilt; that last point is what the multi-dim/product-iterator
subexpression tests exercise.

**Evidence.** Full suite on this branch: matches the pre-change baseline
exactly. Downstream, COPSBenchmark's `robot` and `glider` (eight chained
subexpressions) build as recipes with their original `@add_expr` formulations
and remain bit-identical to `main`'s constructors, including COO order; both
compile through `compile_library` and match their in-Julia models at sizes the
library was never compiled at.

Corroboration for 97fc8787 from a second package: on COPSBenchmark the
map-with-closure path failed for *every* model — its cores carry named refs, a
heterogeneous NamedTuple, so the closure cannot specialize regardless of core
size. With the unroll, all seventeen models compile into one shared library
(16.2 MB, 0 verifier errors) and read back exact. Same fix, independent
trigger: heterogeneity, not scale.

---

The second commit is a small simplification to `_owning_package`: an
extension's `pkgdir` is already its parent package's directory, so reading
the name and uuid from the project file there covers the package and the
extension in one path, and drops the `locate_package` probe (a Base internal)
that existed only to tell the two apart. The extension case stays covered by
the fixture's `RecipeKernelsExaModels` leg.

Issue #284, opened 2026-06-16
Register macros have incorrect hygiene


odow:
You cannot assume that symbols are in scope in the macros. The fix is the interpolate them in various places, so `$ExaModels` and `$Constant`.

```julia
julia> import ExaModels as EM

julia> f(x) = exp(x)
f (generic function with 1 method)

julia> EM.@register_univariate(f, f, f)
ERROR: UndefVarError: `ExaModels` not defined in `Main`
Suggestion: check for spelling errors or missing imports.
Hint: ExaModels is loaded but not imported in the active module Main.
Stacktrace:
 [1] top-level scope
   @ ~/.julia/packages/ExaModels/KqkPA/src/register.jl:59

julia> import ExaModels

julia> EM.@register_univariate(f, f, f)
ERROR: UndefVarError: `Constant` not defined in `Main`
Suggestion: check for spelling errors or missing imports.
Hint: a global variable of this name also exists in ExaModels.
Stacktrace:
 [1] top-level scope
   @ ~/.julia/packages/ExaModels/KqkPA/src/register.jl:63

julia> using ExaModels: Constant

julia> EM.@register_univariate(f, f, f)
```

PR #310, opened 2026-08-13
Fix register macro scoping

pulsipher:
Lack of `ExaModels` prefix on `Constant` leads to scoping errors when `ExaModels.Constant` is not exported to the global scope. This is needed for InfiniteExaModels.


odow:
Closes #284

@pulsipher  the work-around is
https://github.com/lanl-ansi/MathOptAI.jl/blob/b748142866f9f31cb7c51ba70e11ddd6da02f4d5/ext/MathOptAIExaModelsExt.jl#L13

PR #311, opened 2026-08-14
Rename ExaModelsC to ExaModelsCompiler; publish a model's named blocks

sshin23:
Three changes that belong together, since they touch the same files.

## Rename: `ExaModelsC` → `ExaModelsCompiler`

`ExaModelsC` sits one character from `ExaModels`, inside General's AutoMerge name-similarity cutoff (Damerau-Levenshtein 1, cutoff 2; lowercase cutoff 1) — registrable only with a maintainer waiver, as `QiskitIBMRuntimeC` needed. `ExaModelsCompiler` clears the check and says what the package does; the rest of the family already sits well clear (`MadNLPGPU` 3, `ExaModelsPower` 5). Git records it as a rename, so the diff is the real delta rather than a re-add.

## Named blocks, published and queryable

`@add_var(c, x, ...)` registers `x`, and until now a consumer holding only a compiled library had no way to learn that. Each model now publishes its named variable, constraint and parameter blocks — kind, offset, length and dims — through `P_nblocks` / `P_block` / `P_block_name`, plus `P_get_value` / `P_set_value` for parameters. The layout is **instance state**: a recipe instantiated at another size reports that size's offsets, which is why every accessor takes an id.

In ExaModels itself:

```julia
get_vars(m)          # (x = Variable, y = Variable)   — named blocks, by kind
get_vars(m, :x)      # one of them; get_cons / get_pars likewise
```

A name of the wrong kind names the accessor that would work; a typo lists what exists — the blocks share one flat namespace, so both mistakes are easy and want different fixes.

The consumers mirror it exactly ([CNLPModels.jl](https://github.com/madsuite-org/CNLPModels.jl), [cnlpmodels](https://github.com/madsuite-org/cnlpmodels-py), both already on master), so a compiled model is addressed the way an in-Julia one is:

```julia
solution(result, get_vars(m, :pg))     # reshaped to the block's own dims
multipliers(result, get_cons(m, :balance))
set_value!(m, get_pars(m, :load), new_demand)
```

Named blocks are an optional ABI surface: a library that publishes none — a hand-written C library, say — reports empty tuples and is consumed exactly as before.

## Parameters: one vocabulary, and a model owns its values

`set_parameter!` is gone; `set_value!` is the only spelling, paired with `get_value`, matching the two-stage methods that already used those names. Core-level methods were the ones missing and are now present.

**A model copies the core's parameter vector when it is built.** `instantiate` passes a placeholder-free container through by identity, so without the copy every model built from one core shared one `θ` — setting a parameter on one compiled instance set it on all, and two instances at different sizes would have shared one vector. The suite's assertion that instances are independent is what would catch losing this again.

The consequence is deliberate and user-visible: after a model is built, updates go to the **model**, not the core.

```julia
m = ExaModel(c)
set_value!(m, θ, [200.0, 1.0])   # not set_value!(c, θ, ...)
```

The parameters doc page and the "Modify after build" testset were the two places telling the old story and now tell this one. Calls that precede the build are unchanged, since the copy takes the values as they stand. What this buys: several models from one core, each with its own parameters — and several compiled instances in one library likewise.

Also dropped: the schema's ABI version field. Versioning an interface buys nothing before the packages are registered.

## Verification

| suite | result |
|---|---|
| ExaModelsCompiler | 236 / 236 (incl. a five-model library exercising all four instantiation surfaces, per-instance layout at unseen sizes, and a parameter write reaching evaluation) |
| CNLPModels.jl | passed, named blocks 20/20 |
| cnlpmodels-py | 31 passed, 1 skipped |
| ExaModels | re-running at the time of writing — the previous run was 1613/6, all six the "Modify after build" workflow this branch's last commit fixes. **Do not merge before it reports green**; I will post the number here. |


sshin23:
The ExaModels suite is green on the branch tip: **1619 passed, 0 failed, 2 broken** (26m45s) — matching main's own baseline exactly.

The six failures on the previous run were all the `Modify after build` workflow, which updated parameters through the *core* after the model existed. That is the one behavior the per-model parameter copy deliberately ends; the testset and the parameters doc page now update through the model, and calls that precede the build are unchanged.

Full gate set for this branch:

| suite | result |
|---|---|
| ExaModels | 1619 / 0 / 2 broken |
| ExaModelsCompiler | 236 / 236 |
| CNLPModels.jl | passed (named blocks 20/20) |
| cnlpmodels-py | 31 passed, 1 skipped |

The two consumers are already on their masters, so this branch is the last piece.

PR #312, opened 2026-08-14
compile_all: a template for compiling everything a package provides

sshin23:
Stacked on #311 (base is that branch) so it can land after it, or fold in if you'd rather.

`ExaModelsCompiler` declares the verb and owns the keyword contract; a package that ships models implements it for itself in an extension, so **providing models never costs a dependency on a compiler toolchain** — a benchmark set is a modelling package, and juliac is not something you should acquire by loading one.

```julia
compile_all(LuksanVlcekBenchmark; path = "@lvb", sizes = 1000)
```

Callers name the package; providers implement `compile_all(::Val{TheirModule}; …)`. `Val` is the uniform handle — every provider spells it the same way, and it works for a package that has no marker type of its own (ExaModelsPower has none today). The module method is a one-line wrapper over it.

One library rather than one per model, because sizes are deferred: the models share a runtime, one compile amortises across all of them, and a consumer selects with `CNLPModel(lib, :bearing, n)`.

## Verification

Against **LuksanVlcekBenchmark** on its recipes branch, with a draft extension over its eighteen models: compiled `:rosenrock` and `:broyden_banded` into one library, instantiated both at a size never compiled (37), **exact against the in-Julia models** on objective and gradient. ExaModelsCompiler suite **246/246**.

The docstring carries a complete worked extension (weakdeps stanza, ext module, the argument-function spelling for data-defined models) so a provider implements it by copying. @Ohm and @Nabla are testing it against ExaModelsPower and COPS/LVB respectively — emp is the interesting case, since its models take an argument function rather than sizes.

### DIFF
--- a/ExaModelsCompiler/src/ExaModelsCompiler.jl
+++ b/ExaModelsCompiler/src/ExaModelsCompiler.jl
@@ -1744,7 +1744,7 @@ end
 # ── Compiling everything a package provides ──────────────────────────────────
 
 """
-    compile_all(SomePackage; path = "@name", only = nothing, exclude = (), kwargs...)
+    compile_all(SomePackage; path = "@name", kwargs...)
     compile_all(Val(SomePackage); ...)
 
 Compile every model a package provides into **one** shared library.
@@ -1756,7 +1756,6 @@ and its keywords mean the same thing everywhere:
   - `path` — where the library goes, as [`compile_library`](@ref) takes it
     (`"@cops"` installs on the `CNLPMODELS_PATH`); each package picks its own
     default name.
-  - `only` / `exclude` — model names to keep or drop; see [`select`](@ref).
   - everything else is forwarded to `compile_library` (`bundle`, `trim`, …).
 
 One library rather than one per model: sizes are deferred, so the models share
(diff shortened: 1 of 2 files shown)


sshin23:
Two commits since the last review state:

- `a23e698d` drops `select` — a two-line filter does not earn hub API; providers filter their own lists.
- `4d367dc8` makes the emitted library self-describing, answering the two questions a caller holding only a path could not ask: **`cnlp_nmodels` / `cnlp_model_name`** (the catalogue — the only fixed-name symbols, since every per-model entry point needs a prefix and this is where a prefix comes from) and **`P_argtype`** (what a model instantiates from, as a typed signature with optional descriptions — `"int|size"`, `"string|…"`, or a structured model's full field list; every model answers, schema or not, and the signature identifies the entry point without symbol probing).

Gate on the tip: **242/242**, including signature assertions for all four instantiation surfaces.

Consumer counterparts are on both masters (CNLPModels.jl `d4070e5`, cnlpmodels-py `45d3f08`): `available_models`, `argtype`, `schema` answering `nothing` instead of throwing for models without one — plus, Julia-side, the two standard-API methods `ExaModel` had and `CNLPModel` lacked: `cons_nln!` and `hprod!` (assembled COO Hessian-vector product, the piece matrix-free solvers need).

sshin23:
This merge landed on `anvil/rename-compiler` after #311 had already carried that branch into main, so none of it reached main — and the catalogue/argtype commit was pushed after the merge snapshot. Superseded by #313, which is the same three commits cherry-picked onto current main and re-gated there (242/242).

PR #313, opened 2026-08-14
compile_all, a library catalogue, and argument signatures

sshin23:
Supersedes #312, whose content never reached main: it was stacked on `anvil/rename-compiler` and merged into that branch **after** #311 had already carried the branch into main — so the merge landed on a closed base. (Its third commit had additionally been pushed after GitHub took the merge snapshot.) This is the same three commits, cherry-picked onto current main and re-gated there.

**`compile_all`** — ExaModelsCompiler declares the verb and its keyword contract; a package that ships models implements `compile_all(::Val{TheirModule}; …)` in an extension, so providing models never costs a dependency on a compiler toolchain. Callers name the package — `compile_all(LuksanVlcekBenchmark; path = "@lvb", sizes = 1000)` — and the module method forwards to the `Val` form, which is the uniform handle across providers (it works for a package with no marker type of its own). No `select` helper: providers filter their own model lists.

**The library says what it carries** — `cnlp_nmodels` / `cnlp_model_name`, the ABI's only fixed-name symbols, because every per-model entry point needs a prefix and this is where a prefix comes from.

**Every model says what it instantiates from** — `P_argtype` publishes a typed signature, `"type|description,…"`: `""` for a fixed model, `"int|size"`, `"string|…"` for an argument-function model, or a structured model's full field list (`"int|arg1,Vector{f64}|v0,Table{i::int w::f64 s::f64}|bus"`). Every model answers — including those with no builder schema — and the signature identifies the entry point without symbol probing.

Consumer counterparts are already on both masters (CNLPModels.jl `d4070e5`, cnlpmodels-py `45d3f08`): `available_models`, `argtype`, `schema` answering rather than throwing.

Gate on this exact tip: **242/242**, including signature assertions for all four instantiation surfaces; `compile_all` end-to-end verified against LuksanVlcekBenchmark's recipes branch (two models compiled into one library, exact at a size never compiled).


sshin23:
One commit added: the emitted library's contract now has a name — the module docstring and README cite **cnlp ABI v0.1** ([madsuite-org/cnlp-abi](https://github.com/madsuite-org/cnlp-abi), the header-is-the-spec repo, tagged) — plus four stale `MadNLP`-org links corrected. Both consumers already conform to and cite the same tag on their masters, with the spec header vendored into their fixtures so signature drift fails compilation (which caught a real `char*`/`uint8_t*` drift on first compile). Gate on the tip: **242/242**.

With this, merging #313 completes the constellation on ExaModels' side: producer, spec, and both consumers referencing one versioned contract. LVB #4, COPS #16 and ExaModelsPower #57 remain the benchmark-side pieces.

sshin23:
Final state of the branch, after two rounds of field testing by the benchmark providers:

- `7383f516` — **the data-axis fix**: `_length(::AbstractArray)` in ExaModels. A constraint iterating a data collection (catmix's collocation nodes) forms an array axis in its block's size field; `size`/`total`/`multipliers` threw on it, and the compiled layout query degraded that to status 2, leaving six of COPS's seventeen models unconstructable through the consumer. Verified on the reproducer (catmix instantiates, dims `(20,3,2)`, objective exact vs in-Julia); regression tests at both layers. Gates: ExaModels **1621/0/2**, EMC **252/252**.
- `docs commit` — `select`'s docstring carries the two shapes both adopting providers misread (pairs not names; `exclude = ()`).

**Open design item, deliberately NOT in the docs** (its author falsified the drafted guidance before it shipped): a recipe that bakes a dimension (multi-period OPF's bus count) produces a library that silently accepts mismatched inputs — a named argfun cannot carry the frozen value, and a closure that could is refused by the reachability guard. Candidate fixes: admit parameterized callable types as argfuns (`Pkg.MPArgs{14}()` is statically nameable), or check at instantiation with the baked value carried in the core. Neither is in this PR; disclosure via the argtype description is the interim mitigation under discussion.

Field verification on this branch: emp compiles clean on the tip (188s, 0 verifier errors, three prefixes); COPS's closing 17-model run is in flight.

PR #319, opened 2026-08-16
codecov: informational statuses — numbers without CI redness

sshin23:
The `codecov/project` and `codecov/patch` checks keep reporting coverage and its delta on every PR, but `informational: true` means they never fail the CI rollup — a coverage drop is information, not a broken build. Same file going onto MadNLP via #638 and the consumer masters.


sshin23:
Also carries the one stale URL the sweep found here: the README's MadNLP link onto madsuite-org.

